PR #2, opened 2016-06-03
modal component

ilyakamens:


common.styl
```diff
@@ -61,3 +57,4 @@ disabled = '&:disabled,&[disabled]'
 *:focus
   outline 0
 
+border-radius-modal = 6px
```
jbwyme: let's put this in a file called `modals.styl`

components/modals/modals.js
```diff
@@ -0,0 +1,26 @@
+import WebComponent from 'webcomponent';
+
+import css from './modals.styl';
```
jbwyme: AFAICT, this won't work without a stylus plugin for babel (which I'm working on right now)

components/modals/modals.js
```diff
@@ -0,0 +1,26 @@
...
+
+document.registerElement('mpc-modal', class MPCModal extends WebComponent {
```
jbwyme: use `mp` instead of `mpc`

components/modals/modals.styl
```diff
@@ -0,0 +1,14 @@
+@import '../../common'
+
+.mpc-modal
```
jbwyme: `.mp-modal`

examples/index.html
```diff
@@ -9,5 +9,13 @@
     <body>
         <mpc-button type="primary" style="position: absolute; top: 20px;">Save</mpc-button>
```
jbwyme: update to be `mpc`, also you should merge master into your branch

common.styl
ilyakamens: i put this here so that it was available in `buttons.styl` for modal buttons. if i put it in `modals.styl` then i need to import `modals.styl` in `buttons.styl`. you still want me to move it?

examples/index.html
ilyakamens: master is in my branch

common.styl
jbwyme: just have separate variables (or if it's only used once in each, don't use a variable)

PR #3, opened 2016-06-03
update styles to match styleguide

jbwyme:


common.styl
```diff
@@ -1,56 +1,63 @@
...
+/* Text styles */
 
 font-list-item()
-  color color-med-dark
-  font-family font-family-list-item
-  font-size font-size-list-item
-  font-weight font-weight-medium
+  font-family Helvetica Neue Medium
+  font-size 13px
+  line-height 1.7
```
ilyakamens: is this `line-height` correct?

jbwyme: it's literally what zeplin says. My vote is we go by Zeplin and have the
designers ensure it's accurate in the PSDs

jbwyme: also we can make changes to that as needed, if it's wrong now, nbd.

PR #4, opened 2016-06-04
add report and data utils

tdumitrescu:
reorganize into modules for components, report, util


tdumitrescu:
this will have conflicts with your other branches currently out, but it shouldn't be too bad

report/index.js
```diff
@@ -0,0 +1,8 @@
+import MPApp from './mp-app';
+import Persistence from './persistence';
+
+export {
+  MPApp,
+  Persistence,
+};
+export * from './parent-frame';
```
jbwyme: cool didn't know you could do multiple exports

PR #5, opened 2016-06-06
Style updates

jbwyme:


components/buttons/buttons.js
```diff
@@ -3,13 +3,11 @@ import WebComponent from 'webcomponent';
...
-  createdCallback() {
+  attachedCallback() {
     this.root = this.createShadowRoot();
     this.main = document.createElement('div');
+    this.main.className = this.className;
     this.main.classList.add('mp-button');
-  }
-
-  attachedCallback() {
     this.getAttribute('type').split(' ').forEach(type => this.main.classList.add(type));
```
ilyakamens: we can probably remove this and use `class="..."` with what you added on line 9

ilyakamens: in which case can you update the examples too

components/buttons/button.jade
```diff
@@ -0,0 +1,9 @@
+-
+  const attrs = {};
+  if ($component.isAttributeEnabled('disabled')) {
+     attrs.disabled = true;
+  }
```
tdumitrescu: please put this in a helper rather than this weird code chunk in the template

components/buttons/button.styl
```diff
@@ -0,0 +1,76 @@
...
+    &.on-blue
+      background-color blue-300
+      border-color blue-700
+      box-shadow 0px 1px 1px 0px rgba(0, 0, 0, 0.06)
+      color #fff
+
+      &:hover
+        border-color blue-900
+
+
```
tdumitrescu: ilya sez: "extra line!"

PR #8, opened 2016-06-08
add constant $all_events

chzhcn:
for IRB

### DIFF
--- a/lib/report/constants.js
+++ b/lib/report/constants.js
@@ -251,6 +251,7 @@ export const COUNTRIES = {
 };
 
 export const EVENTS = {
+  $all_events: 'All Events',
   $campaign_delivery: 'Notification Sent',
   $campaign_open: 'Notification Opened',
   $campaign_bounced: 'Notification Bounced',


ilyakamens:
LGTM except according to @tdumitrescu version bumps should be a separate commit

chzhcn:
split the commit.

package.json
```diff
@@ -1,7 +1,7 @@
 {
   "name": "mixpanel-common",
   "main": "build/index.js",
-  "version": "0.3.5",
+  "version": "0.3.6",
```
tdumitrescu: minor PSA: in general, it's not a good practice to bump versions in pull requests. whoever publishes the new version will take care of that (in whatever order different branches end up getting merged)

tdumitrescu:
like if https://github.com/mixpanel/mixpanel-common/pull/7 gets merged before this one, someone's just going to have to undo your version change anyway

package.json
chzhcn: Makes sense. Removed that commit.

PR #9, opened 2016-06-08
add utils from analytics to mixpanel-common

crmorford:
moving select functions to mixpanel-common since they're being used in multiple places in collect-everything

context: https://github.com/mixpanel/analytics/pull/7098/files/80c093c42993f5986197ae89cd9b53f3531fb92a#r66316980

### DIFF
--- a/lib/util/index.js
+++ b/lib/util/index.js
@@ -120,7 +120,7 @@ export function objToQueryString(params) {
 }
 
 export function parseUrl(url) {
-  var parser = document.createElement('a');
+  const parser = document.createElement('a');
   parser.href = url;
   return {
     host: parser.host,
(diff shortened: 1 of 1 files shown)


lib/util/index.js
```diff
@@ -115,6 +115,20 @@ export function objectFromPairs(pairs) {
   return object;
 }
 
+export function objToQueryString(params) {
+  return Object.keys(params).map(k => [k, encodeURIComponent(params[k])].join('=')).join('&');
+}
+
+export function parseUrl(url) {
+  var parser = document.createElement('a');
```
ilyakamens: `const`

ilyakamens:
LGTM%C

PR #11, opened 2016-06-16
added some utility functions for testing type of and formatting numbers.

chzhcn:


lib/util/index.js
```diff
@@ -82,6 +82,34 @@ export function commaizeNumber(number, no_conversion) {
...
+}
+
+// if given a floating point number, return a floating point number rounded to the given precision,
+// otherwise return directly.
+export function formattedNumber(number, precision=2) {
+  if (isFloat(number)) {
+    return Number(Math.round(number + 'e' + precision) +'e-' + precision);
+  }
+
+  return number;
```
tdumitrescu: couldn't this entire function be:
```javascript
export function formattedNumber(number, precision=2) {
  return Number(Number(number).toPrecision(precision));
}
```

with one less coercion if you expect `number` always to be a number

lib/util/index.js
```diff
@@ -82,6 +82,34 @@ export function commaizeNumber(number, no_conversion) {
...
+}
+
+// This function calls formattedNumber and converts the return value to string.
+export function formattedNumberStringify(number, precision=2) {
```
tdumitrescu: how come `formattedNumber` isn't sufficient by itself?

lib/util/index.js
```diff
@@ -82,6 +82,34 @@ export function commaizeNumber(number, no_conversion) {
   return commaized;
 }
 
+function isInt(n){
```
tdumitrescu: this is never used

lib/util/index.js
```diff
@@ -82,6 +82,34 @@ export function commaizeNumber(number, no_conversion) {
...
+    return Number(n) === n && n % 1 === 0;
+}
+
+function isFloat(n){
+  return Number(n) === n && n % 1 !== 0;
+}
+
+// if given a floating point number, return a floating point number rounded to the given precision,
+// otherwise return directly.
+export function formattedNumber(number, precision=2) {
```
tdumitrescu: a-z move this after `export function extend`

lib/util/index.js
Quuxplusone: also, nit: plz `) {` not `){`

lib/util/index.js
Quuxplusone: > 1e100 + 'e10'
    "1e+100e10"

lib/util/index.js
```diff
@@ -82,6 +82,34 @@ export function commaizeNumber(number, no_conversion) {
   return commaized;
 }
 
+function isInt(n){
+    return Number(n) === n && n % 1 === 0;
+}
+
+function isFloat(n){
```
Quuxplusone: I'm skeptical that this function could ever be useful. All JS Numbers are `double` by definition (even if they happen to be something-point-zero); but there are some `double`s that could fail this function's check (e.g. infinity, NaN). If you want to check `isfinite(n)`, I'd name this function that, explicitly.

Likewise above, `isInt(n)` is often true for numbers that aren't round-trippable through 32-bit `int`.

lib/util/index.js
chzhcn: I wanted it to only format floating point numbers and leave integer and non-numbers untouched. Maybe I should give it a better name. You can take a look at the other thread for its usage in IRB

chzhcn: https://github.com/mixpanel-platform/irb/pull/30/commits/b948b3774cebcfe8ab64b5ce6ff6ead9c5b2d6b4

jbwyme:
do we still want to merge this or close it?

jbwyme:
bump

PR #13, opened 2016-06-17
"width" and "closeable" attributes for modal

jbwyme:
Also re-factored styles to position modal using flexbox


lib/components/modals/modal.jade
```diff
@@ -1,7 +1,15 @@
...
-    .top-container
-      content(select='[slot-body]')
-    .button-container
-      content(select='[slot-button]')
+    - const style = {}
+    if $component.getAttribute('width')
+      - style.width = $component.getAttribute('width')
+    .modal(style=style)
+      if $component.isAttributeEnabled('closeable')
+        .close-btn(onclick=$helpers.dispatchCloseEvent)
```
tdumitrescu: so the idea here is that whoever inserted the `<mp-modal>` element is responsible for making it disappear when the modal says "i want to close"? that seems like this component is shuffling too much responsibility onto the caller. i'd expect `<mp-modal>` to relieve me of the following responsibilities:
- showing/hiding modal with animations
- dismissing modal when an action button is pressed
- dismissing modal when background is clicked
- dismissing modal when escape is pressed

this is not to say that it shouldn't be firing an "I'm closing" event so the app can react (including if necessary removing the element from the dom entirely), but the actual work of showing/hiding/removing the overlay etc should happen inside the component.

lib/components/modals/modal.jade
```diff
@@ -1,7 +1,15 @@
...
+          .close-icon
+           include ./close.svg
```
tdumitrescu: indentation

lib/components/modals/modal.js
```diff
@@ -4,12 +4,19 @@ import template from './modal.jade';
 
 import css from './modal.styl';
 
+const CLOSE_EVENT = new Event('close');
```
tdumitrescu: - i don't think we should recycle the same Event object for every modal. call `new Event` from within the close handler
- do html elements respond to `onclose`? if not, you won't be able to catch this event with `virtual-dom`; i.e. in the example above, will it work if you replace your `addEventlistener` line with: `modal.onclose = function() {`?

lib/components/modals/modal.styl
```diff
@@ -1,34 +1,60 @@
 @import '../../../stylesheets/common'
 
 .mp-modal
-  pointer-events auto
-  z-index 1
+  position fixed
+  z-index 10
```
tdumitrescu: z-index 1000000000

tdumitrescu: j/k

lib/components/modals/modal.styl
```diff
@@ -1,34 +1,60 @@
...
-      content::content
-        .mp-modal-title
-          font-title()
-        .mp-modal-subtitle
-          font-subtitle()
-          margin 20px 0
+    display flex
+    height 100%
+    justify-content center
+    align-items center
```
tdumitrescu: how easy is this shit these days

lib/components/modals/modal.jade
jbwyme: My thought process was that since this is used with panel, closing the modal will require an app state change regardless. Otherwise it'll just pop back up on the next state update.

tdumitrescu: Sure, the app should be able to hook an event. But:
- why would the modal pop back up? if its declaration in the dom hasn't changed (`<mp-modal closable etc>`) then vdom will just leave it alone
- i feel kinda strongly that these generic components should be usable without `panel`. i.e. stick the element on the page however your app wants to, listen for events from it, and you're good. which this implementation _does_ do, except for the "you're good" part, because the hosting app has to think about the dismissal mechanism.

jbwyme: Does the dom diff not look into the children of the component also? Like
the contents of `mp-modal` will be gone in the current dom but the dom diff
would contain the body wouldn't it?

tdumitrescu: I'm not sure if this is exactly what you're asking, but no, the vdom differ has no way of looking inside a web component; it's based purely on what you declare to it. so if you give it this jade:
```jade
.foo
  mp-modal(closable)
    .some-inserted-content
```
hyperscript/virtual-dom will translate that basically into 3 nested vnodes. in the actual dom, the `mp-modal` custom element inserts a whole bunch of other dom nodes within itself, but virtual-dom doesn't know that, it's not like polling the actual dom for changes. so the next time you give virtual-dom an update to diff, unless one of those 3 vnodes has changed in your updated version, it won't touch the dom there.

lib/components/modals/modal.styl
```diff
@@ -1,34 +1,66 @@
...
+        position absolute
+        right 15px
+        top 15px
+
+        .close-icon svg
+          width 12px
+          .x
+            fill grey-200
+
+        &:hover
```
jbwyme: could single line these two lines

lib/components/modals/modal.jade
```diff
@@ -1,7 +1,16 @@
...
+    .backdrop(onclick=$helpers.backdropClicked)
   .modal-container
-    .top-container
-      content(select='[slot-body]')
-    .button-container
-      content(select='[slot-button]')
+    - const style = {}
+    if $component.getAttribute('width')
+      - style.width = $component.getAttribute('width')
+    .modal(style=style)
```
tdumitrescu: how about pulling 5-7 into a helper, keep the template a little cleaner and high-level

lib/components/modals/modal.jade
```diff
@@ -1,7 +1,16 @@
...
+      if $component.isAttributeEnabled('closeable')
+        .close-btn(onclick=$helpers.closeClicked)
+          .close-icon
+           include ./close.svg
```
tdumitrescu: indentation

lib/components/modals/modal.js
```diff
@@ -10,6 +10,27 @@ document.registerElement('mp-modal', class extends Component {
...
+          this._closeModal();
+        },
+        getType: () => {
+          return this.getAttribute('modalType') || 'modal';
+        }
+      }
     };
   }
+
+  _closeModal() {
```
tdumitrescu: why make this pseudo-private? seems like a perfectly sensible method to have in the component's public api

lib/components/modals/modal.js
```diff
@@ -10,6 +10,27 @@ document.registerElement('mp-modal', class extends Component {
...
+        backdropClicked: () => {
+          if (this.isAttributeEnabled('closeable')) {
+            this._closeModal();
+          }
+        },
+        closeClicked: () => {
+          this._closeModal();
+        },
+        getType: () => {
+          return this.getAttribute('modalType') || 'modal';
```
tdumitrescu: let's stick to dashed names for html attrs, `modal-type`

PR #14, opened 2016-06-17
disable clicks when button is disabled

jbwyme:


lib/components/buttons/button.js
```diff
@@ -21,4 +21,13 @@ document.registerElement('mp-button', class extends Component {
       useShadowDom: true,
     };
   }
+
+  createdCallback() {
+    super.createdCallback(...arguments);
+    this.el.addEventListener('click', (e) => {
```
tdumitrescu: if it works, a cleaner solution would be to make this an `onclick` handler on your wrapper div (the top level el in the jade template), that way you don't have to worry about manually adding and removing handlers. if for some reason you have to do it this way, then add the listener in the `attachedCallback` and remove it in `detachedCallback`

PR #18, opened 2016-06-26
add util/ and report/ symlinks

tdumitrescu:
so you can import from `'mixpanel-common/util'` and `'mixpanel-common/report'`.
if you need the original ES2015 src you can always specify `'mixpanel-common/lib/util'`

### DIFF
--- a/report
+++ b/report
@@ -1,0 +1,1 @@
+build/report/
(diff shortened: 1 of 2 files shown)


jbwyme:
lgtm

PR #20, opened 2016-06-28
download post API query results from browser

jordanmnunez:


lib/util/index.js
```diff
@@ -82,10 +82,69 @@ export function commaizeNumber(number, no_conversion) {
...
+  let parameters = {
+    script: script.replace(/\r/g, '').replace(/\n/g, '\r\n'),
+    download_file: `${filename}.${format}`,
+    params: params || {},
+  };
+  if (format === 'csv') {
+    parameters.format = 'csv';
+  }
+
+  const query = window.MP.api.getQueryOptions('/api/2.0/jql/', parameters, {type: 'POST'});
```
jordanmnunez: should I do anything here to ensure that `MP.api` exists on the window?

lib/util/index.js
```diff
@@ -82,10 +82,69 @@ export function commaizeNumber(number, no_conversion) {
   return commaized;
 }
 
+
+let downloadIdCounter = 0;
+
+export function downloadJQLQuery(script, filename, params=null, format='csv') {
```
tdumitrescu: - this seems more appropriate for `report/`, perhaps as its own module or else just in util
- let's make it `downloadPostData` or even `downloadData` and remove the jql-specific code

lib/util/index.js
tdumitrescu: i'd probably make the caller pass in all the options; see the other comment about making this a generic download util rather than tightly coupled to mp internals

lib/util/index.js
```diff
@@ -82,10 +82,69 @@ export function commaizeNumber(number, no_conversion) {
...
+  // prepare form data
+  let parameters = {
```
ilyakamens: `const`

ilyakamens: same with `exportFrame` and `postForm`

lib/util/index.js
jordanmnunez: sounds good

lib/util/index.js
jordanmnunez: I favoured making this a generic util: ddd3b5c

lib/util/index.js
```diff
@@ -82,10 +82,54 @@ export function commaizeNumber(number, no_conversion) {
...
+  document.body.appendChild(exportFrame);
+
+  // submit request
+  const postForm = document.createElement('form');
+  postForm.action = endpoint;
+  postForm.method = method;
+  postForm.style.display = 'none';
+  postForm.target = exportFrameID;
+
+  // paramsAsFormTags should be html form tags:
```
tdumitrescu: this is kind of a crazy api to expect callers to use. it makes this util a lot harder to use and unnecessarily exposes the mechanism underlying the abstraction. I just want to pass a URL and params/options, basically the same as any normal AJAX call, and not worry about the implementation.

jordanmnunez: i think i got carried away `util`ing it out too much.

PR #24, opened 2016-06-30
Immutable splice

evnp:
Adds some array manipulation utilities

### DIFF
--- a/test/util.js
+++ b/test/util.js
@@ -1,6 +1,13 @@
 import expect from 'expect.js';
 
-import { pluralize } from '../lib/util';
+import {
+  pluralize,
+  immutableSplice,
+  removeIndex,
+  removeValue,
+  replaceIndex,
+  insertAtIndex,
+} from '../lib/util';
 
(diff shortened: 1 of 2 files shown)


test/util.js
```diff
@@ -19,3 +26,97 @@ describe('pluralize()', function() {
     expect(pluralize('die', 4, 'dice')).to.equal('dice');
   });
 });
+
+describe('immutableSplice()', function() {
```
tdumitrescu: awesome, love all the testing. Can we have a couple example-based cases for `immutableSplice` to make it super-simple to look at the tests and see how it's supposed to look in practice? something like:
```javascript
it('removes items', function() {
  expect(immutableSplice([1,2,3,4,5], 2, 2)).to.eql([1,2,5]);
});
```
and similar for inserting

evnp: Good call ce58f9e

evnp:
No worries, should be pretty easy s///g

tdumitrescu:
published 0.6.4, merci

PR #26, opened 2016-07-05
add truncateMiddle, add tests

crmorford:


evnp:
Personally not a fan of alphabetizing utils, I think grouping the functions by their purpose will make this file way more understandable. What is gained by alphabetizing when all our editors have search?

crmorford:
@evnp, oh, my bad. i thought we just wanted to alphabetize all functions and this had slipped through a review. I'll move them back!

tdumitrescu:
That being said, a group of 4 or 5 related functions probably deserves some sort of comment block separating it from its surroundings. or its own module (the main `util` module can still do `export * from './splice-utils';` or something like that)

evnp:
Thanks @crmorford, think I just missed that. @tdumitrescu I'm down for that, will add in https://github.com/mixpanel/mixpanel-common/pull/27

jbwyme:
Alphabetizing functions was a thing ilya wanted to do but I agree that grouping them logically is more useful

evnp:
@tdumitrescu https://github.com/mixpanel/mixpanel-common/pull/27/commits/7b3adcec700db3204f8fedb9f26f65affb51839d

jbwyme:
lgtm

PR #30, opened 2016-07-11
Alert modal

crmorford:


crmorford:
So, i did this with two components/templates/js files at first -- then realized i could combine the components + js files. Now thinking I should probably combine the templates too and just pass in a variable.

ilyakamens:
image

missing border radius

examples/index.html
```diff
@@ -31,10 +32,17 @@
       <div class="mp-modal-subtitle" slot-body>Click on any element on your page (ex. the buy button) to create an event and access all historical data for that event.</div>
       <mp-button class="modal" onclick="closePopup();" slot-button>Create your first event</mp-button>
     </mp-modal>
+    <mp-modal id="alertModal" modal-type="popup" style="display: none" closeable alert>
+      <div slot-body>
+        <div class="alert-modal-title">Your event failed to save</div>
+        <div class="alert-modal-subtitle"">If your issue persists please contact our support team at support@mixpanel.com to get help quickly</div>
```
ilyakamens: looks like you've got an extra double quote here

examples/index.html
```diff
@@ -31,10 +32,17 @@
       <div class="mp-modal-subtitle" slot-body>Click on any element on your page (ex. the buy button) to create an event and access all historical data for that event.</div>
       <mp-button class="modal" onclick="closePopup();" slot-button>Create your first event</mp-button>
     </mp-modal>
+    <mp-modal id="alertModal" modal-type="popup" style="display: none" closeable alert>
```
ilyakamens: minor, but in general our style is to use hyphenated values for ids, classes, etc. (i.e. `alert-modal`)

lib/components/modals/modal.jade
```diff
@@ -1,4 +1,4 @@
-.mp-modal
+.mp-modal(class=alertIcon?"alert":"")
```
ilyakamens: missing some spacing here

lib/components/modals/modal.js
```diff
@@ -8,6 +8,9 @@ document.registerElement('mp-modal', class extends Component {
   get config() {
     return {
       css,
+      defaultState: {
+        alertIcon: this.isAttributeEnabled('alert'),
```
ilyakamens: can we add a verb here like `hasAlertIcon`

lib/components/modals/modal.js
```diff
@@ -62,6 +65,9 @@ document.registerElement('mp-modal', class extends Component {
 
   attachedCallback() {
     super.attachedCallback(...arguments);
+    if (this.isAttributeEnabled('alert')) {
+      this.el.querySelector('.mp-modal').classList.add('alert');
```
ilyakamens: i know this pattern is used immediately below, but i don't think you need this. shouldn't the `alert` class be added to `.mp-modal` in the jade template already?

lib/components/modals/modal.styl
```diff
@@ -109,3 +112,33 @@
...
+      margin-right 15px
+      white-space normal
+      width 33px
+
+      svg 
+        height 29px
+        .fill
+         fill mp-red
+
+
```
ilyakamens: extra line. if you add `"ensure_newline_at_eof_on_save": true,` to your Sublime preferences (`command + comma` when in the app) it should ensure there's always one newline at the end of the file

ilyakamens:
LGTM%C!

lib/components/modals/modal.js
tdumitrescu: yes, manipulating the DOM imperatively is pretty much the #1 anti-pattern in virtual DOM style. This needs to be in a template

jbwyme: except animations.....

tdumitrescu: if the idea is that this class gets added after the template has been rendered, then this code will only work on chrome; on other browsers it won't get rendered til the next animation frame. I still 100% think this needs to be done with state updates, not direct dom manipulation.

jbwyme: not disagreeing, just pointing out that animations fuck us on not manipulating the dom imperatively (or through state as a straight up proxy). cassie you can ignore my comments here.

tdumitrescu: State as a proxy for imperative DOM changes might not be philosophically perfect, but it's at least guaranteed to work. Messing directly with elements added by the virtual dom renderer is a recipe for undefined behavior.

PR #31, opened 2016-07-20
prevent double element registration on the document

ilyakamens:


jbwyme:
sick

lib/components/buttons/button.js
```diff
@@ -37,4 +37,9 @@ document.registerElement('mp-button', class extends Component {
     this.el.removeEventListener('click', this._clickHandler);
     this._clickHandler = null;
   }
-});
+}
+
+if (window['mp-common-registered-components']['mp-button'] !== true) {
+  document.registerElement('mp-button', MPButton);
+  window['mp-common-registered-components']['mp-button'] = true;
+}
```
tdumitrescu: was there a reason not to use the constructor inspection suggested here? http://stackoverflow.com/questions/27334365/how-to-get-list-of-registered-custom-elements

it avoids polluting the global namespace.

PR #37, opened 2016-07-25
super important sum() util

tdumitrescu:
implemented without reduce and other trickery, to keep it lightweight and fast enough

### DIFF
--- a/lib/util/index.js
+++ b/lib/util/index.js
@@ -153,6 +153,14 @@ export function pluralize(singular, number, plural) {
   return number === 0 || number > 1 ? plural : singular;
 }
 
+export function sum(arr) {
+  let _sum = 0;
+  for (let i = 0; i < arr.length; i++) {
+    _sum += arr[i];
+  }
+  return _sum;
+}
+
 export function truncateMiddle(string, len) {
   if (string) {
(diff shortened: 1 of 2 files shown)


jordanmnunez:
lgtm 👍

ilyakamens:
LGTM

lib/util/index.js
```diff
@@ -153,6 +153,14 @@ export function pluralize(singular, number, plural) {
...
 }
 
+export function sum(arr) {
+  let _sum = 0;
+  for (let i = 0; i < arr.length; i++) {
+    _sum += arr[i];
+  }
+  return _sum;
+}
+
```
evnp: Why not the traditional `arr.reduce((a, b) => a + b, 0)`? Either way is cool, just curious

evnp: jesus, need to read comments. Ignore me :D

tdumitrescu: haha well another point only in the tests, not github description, is that this implementation works out of the box for non-array objects like `arguments`

evnp: Ah cool, the more you know. As soon as you put reduce and prototype bs together the above starts to look pretty nice in terms of readability

PR #40, opened 2016-08-11
use attributes to fix modal

ilyakamens:


examples/index.html
```diff
@@ -30,17 +30,17 @@
       Mouse over me!
       <mp-tooltip>This is my tooltip</mp-tooltip>
     </mp-button>
-    <mp-modal id="modal" closeable>
+    <mp-modal id="modal" closeable hide>
```
jbwyme: it'd be more consistent to use `closed` instead of `hide` throughout here

PR #43, opened 2016-08-17
more consistent class prefixing

jbwyme:


lib/components/modals/modal.jade
```diff
@@ -1,24 +1,24 @@
-.mp-modal(class={
-  alert: $component.isAttributeEnabled('alert'),
-  absolute: $component.isAttributeEnabled('not-fullscreen'),
-  closed: $component.isAttributeEnabled('closed'),
+.mp-modal-stage(class={
```
ilyakamens: why `stage`?

ilyakamens: you know i hate this word can we use a different one

ilyakamens:
cool LGTM%C

lib/components/modals/modal.jade
jbwyme: if you want to suggest a better name go for it, take a look at the class hierarchy first though

tdumitrescu:
Shipitsquirrel

lib/components/modals/modal.jade
ilyakamens: it's hard for me to suggest a better name because i don't understand what stage is

PR #47, opened 2016-08-23
add style guide graph colors

tdumitrescu:
NB `color-seafoam` is the same as `seafoam`; i'd love to see the old one get removed once autotrack has a chance to update. Using the `color-` prefix on these to avoid collisions with global css colors.

### DIFF
--- a/stylesheets/common.styl
+++ b/stylesheets/common.styl
@@ -16,6 +16,14 @@ mp-red = #e4567b
 mp-yellow = #ffd209
 seafoam = #47b6ac
 
+color-aquamarine = #106eca
+color-navy = #23588c
+color-orange = #f2af34
+color-seafoam = #47b6ac
+color-turquoise = #24d2ef
+color-ultramarine = #106eca
+color-yellow = #ffd209
+
 /* Text styles */
 
 font-list-item()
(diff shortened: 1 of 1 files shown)


zakj:
in the case of collisions, the stylus var would win, right? in that case why care about collisions?

sorry to nitpick I'm just envisioning a world filled with `color: color-foo; background-color: color-bar` and feeling sad

tdumitrescu:
Because i don't want to have to think about what would win, or know when looking at a particular stylesheet that this 'yellow' is going to be default 'yellow' because common styles haven't been imported. It's just more explicit. Happy to use a less sad prefix like `mp-` which seems to be used up above.

zakj:
y u not use `@` prefiks liek less?!1

`mp-` wfm

tdumitrescu:
well because obvs you need the `@` for looking up property names http://stylus-lang.com/docs/variables.html#property-lookup (lolol)

i mean we could use `$yellow` but that might make me shoot myself in the face

tdumitrescu:
https://github.com/mixpanel/mixpanel-common/pull/47/commits/08646f824f2fb5aac8a06dac09883e4364cbfc3e

zakj:
oooooh I totally forgot about that property lookup feature in stylus and I kind of love it.

PR #48, opened 2016-09-13
jjo's sweet new icon set

tdumitrescu:
importable from anywhere


stylesheets/common.styl
```diff
@@ -1,4 +1,4 @@
-/* Color palettte */
+/* Color palette */
```
jordanmnunez: how long as that been bugging you for

tdumitrescu: whatttt

PR #49, opened 2016-09-14
Added the value icons

jayjo:
value_relative.svg and value_absolute.svg added into the repository for values in the extras menu


tdumitrescu:
yes!

PR #51, opened 2016-09-26
Iconography

jayjo:
added in the lightning bolt so ilya doesn't have to draw it by hand


jayjo:
They type-casting icons are a bit different - we may need to discuss how we want to approach styling those some (I know the Boolean for sure) have multiple colors in the SVG that need to change on different states.

tdumitrescu:
whoa old PR, same branch. let me see about cleaning up anything outdated here. for multi-color icons, i'd suggest:
- don't hardcode any fills into the SVG markup
- have some base global CSS that defines the normal color states of those ones (e.g. that would set `#crazy-type-icon path.color1` to `blue-300`)
- in general, i'd prefer to require the least amount of icon-specific knowledge possible from their users. so for instance if i'm using `crazy-type-icon` somewhere in my markup I can try styling it first by CSSing `fill: red` on `#my-icon svg`, then if I need to change specific parts I could select for `#my-icon svg g.color1` rather than `svg g#the-twirl-on-the-left`

tdumitrescu:
merci!

PR #53, opened 2016-10-04
add purple button styles

jbwyme:


examples/index.html
```diff
@@ -76,12 +82,26 @@
...
       });
 
+      var purpleModal = document.getElementById('purple-modal');
+      purpleModal.addEventListener('change', function(e) {
+          if (e.detail.state === 'closed') {
+              purpleModal.setAttribute('closed', true);
+          }
+      });
+      var openPurpleModal = function() {
+        purpleModal.removeAttribute('closed');
```
tdumitrescu: is this something that our modals don't do automatically when `open()` is called?

lib/components/buttons/button.styl
```diff
@@ -124,3 +115,21 @@ background-primary-red-hover = mp-red linear-gradient(to bottom, #ED92A4, #EA849
...
+      border-image-slice 1
+
+      &:hover{not-disabled}
+        background background-primary-red-hover
+
+    &.purple
+      background mp-purple
+      border 1px solid #7858b8
+      box-shadow none
+
```
crmorford: + font-weight: bold;
+ font-size: 14px;

crmorford:
LGTM % comment

examples/index.html
jbwyme: no but it probably should, i'll fix that

lib/components/buttons/button.styl
jbwyme: This is different than our existing modal styles and makes them not-consistent. I'll check with the design team.

jbwyme: We're going to stick with font-size 14px so it's consistent. If we WERE to change it here, we should change it for all modal buttons. I'm also going to update modal buttons to use font-weight bold instead of 400.

examples/index.html
```diff
@@ -76,32 +82,49 @@
...
               alertModal.setAttribute('closed', true);
           }
       });
 
+      var purpleModal = document.getElementById('purple-modal');
+      purpleModal.addEventListener('change', function(e) {
+          if (e.detail.state === 'closed') {
+              purpleModal.setAttribute('closed', true);
+          }
+          });
```
tdumitrescu: weird indentation

examples/index.html
```diff
@@ -76,32 +82,49 @@
...
+      alertModal.addEventListener('change', function(e) {
           if (e.detail.state === 'closed') {
               alertModal.setAttribute('closed', true);
           }
       });
 
+      var purpleModal = document.getElementById('purple-modal');
+      purpleModal.addEventListener('change', function(e) {
+          if (e.detail.state === 'closed') {
+              purpleModal.setAttribute('closed', true);
```
tdumitrescu: i think here too the modal should probably take care of its own attr

lib/components/modals/modal.js
```diff
@@ -45,6 +45,7 @@ class MPModal extends Component {
   }
 
   open() {
+    this.removeAttribute('closed');
```
tdumitrescu: NB this will trigger a re-render through `attributeChangedCallback`. Is that OK?

examples/index.html
jbwyme: hmm ok now I'm thinking there must have been a reason we did it this way, probably some interoperability issues with panel

tdumitrescu: I'm guessing the idea is that the parent should be responsible for managing the `closed` attr entirely. Which doesn't work that great in combination with imperative api methods like `close()` and `open()`.

PR #56, opened 2016-10-12
Use shared eslint

jbwyme:


tdumitrescu:
What would this look like with only the linter updates?

lib/report/mp-app.js
```diff
@@ -1,3 +1,5 @@
+/*eslint camelcase: ["error", {properties: "never"}]*/
```
tdumitrescu: I don't think we should be throwing in a linter override for an entire code module like this (excepting cases like the constants module above). How many lines actually need this?

lib/util/index.js
```diff
@@ -2,101 +2,102 @@ export * from './array';
...
 
   if (neg) {
-    commaized = '-' + commaized;
+    commaized = `-` + commaized;
   }
 
   return commaized;
 }
 
+
```
tdumitrescu: extra blank line

package.json
```diff
@@ -37,7 +37,8 @@
     "babel-polyfill": "6.9.1",
     "babel-preset-es2015": "6.9.0",
     "css-loader": "0.23.1",
-    "eslint": "2.11.1",
+    "eslint": "3.7.1",
+    "eslint-config-mixpanel": "^2.0.0",
```
tdumitrescu: bump to 2.0.1?

tdumitrescu:
Not gonna do this, or reopening with fresh branch?

jbwyme:
gonna do it but it was too out of date so gonna open fresh branch

PR #61, opened 2016-10-24
svg-icon component for inline css-stylable SVG icons

tdumitrescu:
incorporates https://github.com/mixpanel/mixpanel-common/pull/60

usage looks like:
```html
<svg-icon icon="arrow-left"></svg-icon>
```

style guide has been redone as a Panel component because all the html-injection code was getting gross.

guide now looks like:
screen shot 2016-10-24 at 1 01 09 pm

To fix later:
- lightning-bolt vs lighting-bolt (the purple icon needs replaced, but not until existing uses are audited for replacement with the new component)
- `type-boolean` has several paths that need individual coloring but currently have no classes to select on, so the whole thing becomes a filled in square


stylesheets/forms.styl
```diff
@@ -1,21 +1,21 @@
...
+  color grey-700
+  display inline-block
+  font-size 13px
+  font-weight 500
+  transition border-color 150ms ease-out
 
-    ::placeholder
-      color grey-500 !important
-      font-weight 400 !important
+  ::placeholder
```
zakj: I know you're just moving this but doesn't it need to be `&::placeholder`?

wct.conf.json
```diff
@@ -5,8 +5,7 @@
     "local": {
       "disabled": true,
       "browsers": [
-        "chrome",
-        "firefox"
```
zakj: I think we intentionally had firefox in here to test the webcomponents polyfill

.gitignore
```diff
@@ -1,5 +1,6 @@
 .DS_Store
 build
+components
```
tdumitrescu: It comes from https://github.com/mixpanel/mixpanel-common/pull/61/commits/722794e9a3c6607e99e8840170a109f4a5beed7a. Same thing we do for the `report` and `util` modules to make them top-level importable via `mixpanel-common/report` etc - copy the built versions into the package root (since npm doesn't do symlinks).

zakj: ah, got it. maybe `/components` then, to not ignore `lib/components`

wct.conf.json
tdumitrescu: Yeah this is for the local (non-CI) version of tests. Ever since firefox 47 most people haven't been able to get it to start up for them: https://github.com/Polymer/web-component-tester/issues/367

(haha latest comment: "Just ran into this myself. Now that Polymer Summit is over, got a sec to take a look?")

.gitignore
tdumitrescu: Whoa yeah, good catch. That goes for all of them - I guess git will still track changes on gitignored-but-tracked files, since we haven't had any trouble updating `report` and `util` dirs.

tdumitrescu: https://github.com/mixpanel/mixpanel-common/pull/61/commits/f4a705089195f7aba7db9a4c2a98c908c04385c2

stylesheets/forms.styl
tdumitrescu: BTW same with `-webkit-font-smoothing: antialiased;` which is all over the dang place in generated css

jbwyme: yeah it's used in autotrack, it does suck that it gets generated over and over but on the other hand it guarantees that all of our form styles are the same by default (same with the link styles in typography.styl). One way to fix this would be to have two general imports: defaults.styl (which includes explicit style declarations) and common.styl (which includes variables only). thoughts?

tdumitrescu: Yes. `common.styl` should be importable anywhere, which means it should not contain anything but vars and mixins and no top-level styles. `defaults.styl` can be imported once at an app's entrypoint along with polyfills and all the other bootstrapping junk.

zakj: +1

PR #62, opened 2016-11-02
mp-toggle component

tdumitrescu:
@jbwyme @EricBenjamin 

The styling on this bad boy is totes grodes:
screen shot 2016-11-01 at 6 35 50 pm

But the gray/white version matches what's currently in IRB, and @EricBenjamin can style up blue before @ldelbeccaro has need of it.

@jbwyme some implementation notes:
- I extracted that weird c/p'ed component-registration code into a `registerMPComponent()` function in `registration.js`
- reordered widgets in style guide alphabetically
- the `mp-toggle` component itself has an API where you specify the choices as `mp-toggle-option` elements nested under the `mp-toggle` element:
```html
<mp-toggle>
  <mp-toggle-option value="llama">Llamas!!</mp-toggle-option>
  <mp-toggle-option value="capybara">wat</mp-toggle-option>
</mp-toggle>
```
Possibly controversial: rather than trying to style the inserted content elements from the light DOM, which is a sketchy proposition with an unstable API (I think even removed altogether in WCv1), the implementation works by scanning the user's element and mapping `mp-toggle-option` elements into `.mp-toggle-option` elements in the template-generated shadow DOM. The main drawback is that if you want a dynamically-changing choice set, you have to make sure the `mp-toggle` gets rerendered after any change; but 1) I don't see a lot of need for dynamic options, and 2) [MutationObserver](https://developer.mozilla.org/en-US/docs/Web/API/MutationObserver) could make it seamless, if really desired. Reactions welcome.


examples/style-guide/index.jade
```diff
@@ -91,3 +84,25 @@
...
         .mp-modal-alert-subtitle If your issue persists please contact our support team at support@mixpanel.com to get help quickly
+
+  .section.toggles
+    h1 Toggles
+    p One element out of a set of two or more is always selected.
+    mp-toggle.mp-toggle-blue.inline-toggle(
+      attributes={selected: blueToggleValue}
+      onchange=$helpers.blueToggleChanged
+    )
+      mp-toggle-option(attributes={value: 'option1'}) Option 1
```
jbwyme: how do you feel about just using `option` instead of `mp-toggle-option` ?

lib/components/toggles/index.styl
```diff
@@ -0,0 +1,39 @@
+@import '../../../stylesheets/common'
+
```
jbwyme: although i have no idea where the final spec will end up, I've been adding 
```
:host
  display inline-flex
```

so the host element gets the full layout

jbwyme:
lgtm % c

lib/components/toggles/index.styl
tdumitrescu: Indeed, thanks for reminding me to style some host

examples/style-guide/index.jade
tdumitrescu: My main reservation is that it's less flexible, inasmuch as an `<option>` node cannot legally have any children (except text content). Using a custom element (even an unregistered one, as here) means that we could painlessly expand the API, for instance if we wanted to allow embedded `img` tags or `span`s.

lib/components/toggles/index.styl
tdumitrescu: https://github.com/mixpanel/mixpanel-common/commit/815eaac7db977bdc2e5f01eabd978a9d01b4fa7f @jbwyme since it seems you're in the process of making `stylesheets/default` i won't do any more with the `components/default.styl` for now

tdumitrescu:
merged 0.20.0

PR #63, opened 2016-11-02
Styleguide updates

jbwyme:
- Added full color palette
- Updated typography to be a bit more granular
- Removed `font-title`, `font-subtitle`, `font-list-item` as they were not the right level of abstraction
- Updated button styles
- Added support for a handful of icons in buttons


examples/style-guide/index.html
```diff
@@ -1,10 +1,10 @@
-<!DOCTYPE html>
+<!DOCTYPE html PUBLIC "-//W3C//DTD XHTML 1.0 Transitional//EN" "http://www.w3.org/TR/xhtml1/DTD/xhtml1-transitional.dtd">
```
tdumitrescu: ugh what a fiasco. still the case for mixpanel.com, but not any of the modern platform apps...

examples/style-guide/index.jade
```diff
@@ -52,16 +65,29 @@
...
       mp-button.mp-button-primary.mp-button-on-blue primary on-blue
       mp-button.mp-button-primary.mp-button-on-blue(disabled) disabled
     p.button-field
       mp-button.mp-button-secondary.mp-button-on-blue secondary on-blue
       mp-button.mp-button-secondary.mp-button-on-blue(disabled) disabled
 
+    h1 Buttons w/ icons
+    p
+      mp-button.mp-button-secondary(attributes={icon: 'x'}) Close
+      mp-button.mp-button-secondary(attributes={icon: 'logout'}) Exit
```
tdumitrescu: love how easy this api is

lib/components/buttons/button.styl
```diff
@@ -1,120 +1,159 @@
...
       &:hover{not-disabled}
-        background background-primary-red-hover
+        background-image linear-gradient(0deg, red-300 0%, red-300 100%)
 
+      &:active{not-disabled}, &.active{not-disabled}
+        background-image linear-gradient(0deg, mp-red 0%, red-300 100%)
+        box-shadow 0px 1px 1px 0px rgba(0,0,0,0.0), inset 0px 2px 2px 0px rgba(0,0,0,0.2)
+        content-color(rgba(255,255,255,0.6))
+
+    // deprecated
```
tdumitrescu: any actual uses of the purple button? if not let's just delete

lib/components/buttons/button.styl
```diff
@@ -127,6 +166,22 @@ not-disabled = ':not([disabled]):not(:disabled)'
...
 
+    /* ICON STYLES */
+
+    +content-styles()
+      svg-icon
+        &[icon='x']
+          margin-right 2px
+
+        &[icon='plus']
+          icon-size(size: 30px, pos-top: -4px)
```
tdumitrescu: so it's basically the case that neither `plus` nor `plus-large` is the appropriate size by default?

stylesheets/colors.json
```diff
@@ -1,25 +1,56 @@
...
-   "mp-orange":"#f2af34",
-   "mp-purple":"#9271e2",
-   "mp-red":"#e4567b",
-   "mp-seafoam":"#47b6ac",
+   "mp-seafoam":"#46b5ab",
    "mp-turquoise":"#24d2ef",
-   "mp-ultramarine":"#106eca",
-   "mp-yellow":"#ffd209"
+   "mp-ultramarine":"#106eca"
 }
```
tdumitrescu: awesome!

stylesheets/colors.styl
```diff
@@ -1 +1,3 @@
 json('colors.json')
+blue-300 = mp-blue
```
tdumitrescu: any way to do this the other way around? define 100-900 for each color and then add aliases for `mp-<color>`?

stylesheets/common.styl
```diff
@@ -1,4 +1,25 @@
...
+    {block}
+
+host-styles(tagName, additionalSelector='')
+  // apply host styles for FF, Safari, IE
+  // :host doesn't work with webcomponents polyfill, so we have to use the tag name
+  {tagName}{additionalSelector}
+    display inline-flex
+    {block}
+
+  // since :host is a psuedo selector we need to wrap any additional selector with parens
```
tdumitrescu: pseudo

stylesheets/forms.styl
```diff
@@ -1,25 +0,0 @@
...
-  &:focus
-    border-color blue-700
-    transition border-color 250ms ease-in
-
-*:focus
-  outline 0
-
-*::-ms-clear
-  height 0
-  width 0
```
tdumitrescu: yessssss

stylesheets/typography.styl
```diff
@@ -1,60 +1,36 @@
...
+font-paragraph()
+  color grey-700
+  font-size-medium()
+  font-family family-helvetica
+  font-stretch normal
+  font-weight weight-medium
+  line-height 18px
 
-*
-  -webkit-font-smoothing: antialiased;
```
tdumitrescu: love to see this getting cleaned up

lib/components/buttons/button.styl
jbwyme: correct

tdumitrescu: Kinda nuts, since as far as i know this is currently the _only_ use of this icon and it's already non-standard.

lib/components/svg-icon/index.js
```diff
@@ -1,10 +1,22 @@
 import WebComponent from 'webcomponent';
+import '../registration.js';
```
tdumitrescu: delete this

lib/components/defaults.styl
```diff
@@ -1,2 +0,0 @@
-:host
-  display inline-flex
```
tdumitrescu: so the idea is that each component should import common and call `host-styles()` with its own custom mods? a little worried that if you never call it, you'll have a somewhat non-standard component, and that's easier to leave out than a standard import. but it's no biggie

stylesheets/colors.json
```diff
@@ -1,25 +1,56 @@
...
+   "grey-500":"#9cacbb",
+   "grey-300":"#c1ccd5",
+   "grey-200":"#d8e0e6",
+   "grey-150":"#e5eaef",
+   "grey-100":"#eff3f5",
+   "grey-50":"#f9fafc",
+   "white": "#ffffff",
+
+   "green-700":"#19B264",
+   "mp-green":"#2DC376",
```
tdumitrescu: also at some point can we get consistency on hExColOrcAse? just noticed from the tooltips with color codes

lib/components/buttons/button.styl
jbwyme: yeah.... idk man

lib/components/defaults.styl
jbwyme: If there was a way to automatically have this apply to all components including those using webcomponents polyfill i'd do it. However, since this would apply to all native web components but not those using the polyfill it seems preferably to be explicit.

PR #65, opened 2016-11-03
mp-toggle fixes for non-chrome browsers

tdumitrescu:
@jbwyme opened this against yr styleguide branch for the real diff

### DIFF
--- a/lib/components/toggles/index.js
+++ b/lib/components/toggles/index.js
@@ -14,10 +14,12 @@ registerMPComponent('mp-toggle', class extends Component {
 
       helpers: {
         selectOption: value => this.value = value,
-        toggleOptions: () => Array.from(this.querySelectorAll('mp-toggle-option')).map(el => ({
-          text: el.textContent,
-          value: el.getAttribute('value'),
-        })),
+        toggleOptions: () => Array.from(this.children)
+          .filter(el => el.tagName.toLowerCase() === 'mp-toggle-option')
+          .map(el => ({
+            text: el.textContent,
+            value: el.getAttribute('value'),
+          })),
       },
     };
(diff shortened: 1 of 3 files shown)


lib/components/toggles/index.js
```diff
@@ -14,10 +14,12 @@ registerMPComponent('mp-toggle', class extends Component {
 
       helpers: {
         selectOption: value => this.value = value,
-        toggleOptions: () => Array.from(this.querySelectorAll('mp-toggle-option')).map(el => ({
```
jbwyme: this doesn't work?

lib/components/toggles/index.styl
```diff
@@ -1,6 +1,7 @@
+@import '../../../stylesheets/common'
 @import '../../../stylesheets/defaults'
 
-:host
++host-styles(mp-toggle)
```
tdumitrescu: without the +? tried, didn't work. stylus is so terrible with silent failures

lib/components/toggles/index.js
tdumitrescu: only in native implementation... https://github.com/mixpanel/mixpanel-common/pull/65/commits/3052b74036289b17cdeb18b7872013284ff31be5

lib/components/toggles/index.styl
tdumitrescu: oh you mean without the display stuff. no i need `inline-block` here not `inline-flex`

lib/components/toggles/index.js
jbwyme: 😷

tdumitrescu:
merged to master

PR #67, opened 2016-11-04
Reverted selected state toggle color

EricBenjamin:


EricBenjamin:
^ this thing... I can't even.

PR #71, opened 2016-11-15
list menu dropdown added

rickydoar:
Added a list menu dropdown component


tdumitrescu:
DROPDOWNS

jbwyme:
Based on our slack discussion, I think this is the wrong level of abstraction for what we need. Really we want a "popup" type component that handles hiding, showing, animating, and container styles but has no content. Also given the differences in various menus, it's probably best that we just have `styl`es for the menu items but it doesn't really need to be a component.

tdumitrescu:
Oh you mean like a `pane` component that overlays and is content-agnostic? Talk to @jordanmnunez

jbwyme:
I'd like to be able to replace these pane/popup/whatever components with a single standard mp-common one: 
https://www.dropbox.com/s/f8ggd9fd7cfye72/Screenshot%202016-11-15%2012.51.17.png?dl=0
https://www.dropbox.com/s/7otclcxep3jac0k/Screenshot%202016-11-15%2012.51.50.png?dl=0
https://www.dropbox.com/s/l0w189imsg0oi02/Screenshot%202016-11-15%2013.30.00.png?dl=0

tdumitrescu:
Yup, looks like what @jordanmnunez is on in IRB

EricBenjamin:
From a design standpoint, we were hoping to have: 

1.  Modals (background greyed out / shadow most intense)
3.  Dialogues (background greyed out / shadow most intense)
4.  Pop-up (shadow mid-level intensity)
5.  Drop down (shadow mid-level intensity)

The drop downs could then have component elements like: list, icon list, dual list, search bar, toggle button, button bar

PR #74, opened 2016-11-16
mp-input-group component

jbwyme:
http://cometapp.us/c/oQjt3DW2


lib/components/input-group/index.jade
```diff
@@ -0,0 +1,23 @@
...
+      - inputAttrs.disabled = true
+    input.mp-input-group-input(
+      oninput=$helpers.setValue
+      onkeyup=$helpers.submitIfEnter
+      attributes=inputAttrs
+    )
+    mp-button.mp-button-primary.mp-input-group-btn(
+      onclick=$helpers.submit
+      attributes={
+        disabled: disabled || value === '',
```
tdumitrescu: Should we trim `value` to keep the button disabled for all-whitespace input?

lib/components/input-group/index.jade
```diff
@@ -0,0 +1,23 @@
...
+        'style-override': '.mp-input-group-btn.mp-button-container { border-top-left-radius: 0px; border-bottom-left-radius: 0px; width: 90px; }',
+      }
+    )
+      if $component.isAttributeEnabled('saving')
+        .mp-input-group-spinner
+          include ./spinner.svg
```
tdumitrescu: pretty generic spinner - can we get it into the standard `<svg-icon>` set?

lib/components/input-group/index.jade
```diff
@@ -0,0 +1,23 @@
...
+      else
+        | #{$component.getAttribute('label') || 'Save'}
```
tdumitrescu: interpolating an entire line always seems weird to me. is this different than:
```jade
= $component.getAttribute('label') || 'Save'
```

lib/components/input-group/index.js
```diff
@@ -0,0 +1,64 @@
...
+    return this.state.value;
+  }
+
+  set value(value) {
+    this.update({value});
+  }
+
+  attachedCallback() {
+    super.attachedCallback(...arguments);
+    const updates = {};
```
tdumitrescu: `updates` isn't actually used is it? (shouldn't a linter be picking this up?)

lib/components/input-group/index.js
```diff
@@ -0,0 +1,64 @@
...
+    if (this.getAttribute('value')) {
+      updates.value = this.getAttribute('value');
+      this.update({value: this.getAttribute('value')});
+    }
+  }
+
+  detachedCallback() {
+    super.detachedCallback(...arguments);
+  }
```
tdumitrescu: i guess this can be removed?

lib/components/input-group/index.js
```diff
@@ -0,0 +1,64 @@
...
+
+  attributeChangedCallback(name) {
+    super.attributeChangedCallback(...arguments);
+    if (this.initialized && name === 'value') {
+      this.update({value: this.getAttribute('value')});
+    }
```
tdumitrescu: can extract this and the version in `attachedCallback` into a single method like `updateValueFromAttr()`

lib/components/input-group/index.jade
jbwyme: up to @jayjo

PR #75, opened 2016-11-18
Add a custom transpiler binary for watching and transpiling lib files

jbwyme:


transpile.js
```diff
@@ -0,0 +1,108 @@
...
+          'extensions': ['.svg'],
+        }],
+      ],
+      presets: ['es2015'],
+    };
+    return  babel.transformFileSync(file, options).code;
+  },
+
+  jade: file => {
+    const source  = fs.readFileSync(file).toString();
```
tdumitrescu: twooooooooo space!

transpile.js
```diff
@@ -0,0 +1,108 @@
...
+    const vjade = require('virtual-jade');
+    const vjadeOptions = {
+      filename: file,
```
tdumitrescu: cool, just to make sure, have you tested this with jade templates including other files?

transpile.js
```diff
@@ -0,0 +1,108 @@
...
+      .render()
+      .replace(/\n/g, ' ')
+      .replace(/"/g, /\/"/);
+    return 'module.exports = "' + css + '";\n';
+  },
+};
+
+function transpileFile(file) {
+  const ext = file.split('.').pop();
+  if (['js', 'jade', 'styl'].indexOf(ext) > -1) {
```
tdumitrescu: how about
```js
if (['js', 'jade', 'styl'].includes(ext)) {
```

transpile.js
jbwyme: yeah we have templates that inline svgs in the source already

PR #82, opened 2016-11-29
refactor modal api, add mp-confirm, mp-confirm-delete, mp-button-bar

jbwyme:
- Cleaned up the modal api a bit to make it more consistent
- Cleaned up the button component since it was a bit overloaded
- Added three new components: `mp-confirm`, `mp-confirm-delete`, and `mp-button-bar`

Note: there will need to be changes in `analytics` before updating once this is merged in.


stylesheets/defaults/index.styl
```diff
@@ -3,6 +3,7 @@
 
 *
   -webkit-font-smoothing: antialiased;
+  box-sizing border-box
```
jbwyme: contentious: maybe this should go into a `defaults.styl` file under the `components` dir like we originally tried for `host-styles`.

PR #84, opened 2016-11-30
Jeff new

jayjo:


stylesheets/colors.json
```diff
@@ -46,11 +46,12 @@
...
 
    "seafoam":"#47B6AC",
-   "mp-aquamarine":"#6FE9AA",
-   "mp-ultramarine":"#0F6ECA",
-   "mp-midnight":"#0E3963",
-   "mp-navy":"#23588C",
-   "mp-seafoam":"#46B5AB",
-   "mp-turquoise":"#24D2EF",
-   "mp-ultramarine":"#106ECA"
+   "aquamarine":"#6FE9AA",
```
jbwyme: this will have far reaching effects. Anywhere anyone was using those variables will have to be changed....

stylesheets/defaults/index.styl
```diff
@@ -1,5 +1,6 @@
 @import '../common.styl'
 @import 'svg-icon.styl'
+@import 'styleguide-app.styl'
```
jbwyme: this isn't the right place for this, shouldn't this just be the `index.styl` in your `style-guide-new` folder anyways?

examples/style-guide/index.js
```diff
@@ -11,15 +11,27 @@ import { SVG_ICONS } from '../../build/components/svg-icon';
 import template from './index.jade';
```
jbwyme: this PR should leave all `style-guide/*` files untouched

examples/style-guide/index.jade
```diff
@@ -1,155 +1,177 @@
-.style-guide
```
jbwyme: this PR should leave all style-guide/* files untouched

package.json
```diff
@@ -11,7 +11,8 @@
     "lint": "eslint lib",
```
jbwyme: i know this happened automatically, i need to fix that. Let's remove these changes from this PR though.

stylesheets/colors.json
tdumitrescu: It was also a conscious decision about distinguishing easily between mp-specific stylus vars and css default colors, as well as not overriding defaults; e.g. if you write `background-color navy`, is that browser default navy or mp's version? well, the answer is that it will change depending on whether you've included the stylus mixins. not a great idea.

jayjo: In looking through the repo, most of them are not used anywhere else outside of the `colors.json`. A couple are used in a `.less`. Let's change them, because they shouldn't be and I don't think ever have been named with the 'mp' prefix.

jbwyme:
I think we should keep them. We use them here
https://github.com/mixpanel-platform/dashboard/blob/01bc36627ff6c39eef8e6dd83047003d5264b864/src/utils/colors.js)
and
like Ted said, not prefixing will clash with browser defaults which is just
generally not a good idea.

On Mon, Dec 5, 2016 at 10:57 AM, Jeffrey Jorgensen <notifications@github.com

jbwyme:
http://www.w3schools.com/colors/colors_names.asp

stylesheets/colors.json
jayjo: To your point @tdumitrescu, I think the `$` distinguisher would be enough to differentiate between the two - probably better practice to use it to always know what a variable is anyway.

PR #87, opened 2016-12-01
publish stylus and css from lib/

tdumitrescu:
Most changes are to the build/watch scripts, to apply different transformations to different source dirs:
- stylus files in `lib/stylesheets/mixins` don't get transformed; consumers should pull them in directly to get mixins
- stylus files elsewhere in `lib/stylesheets` get compiled to `.css` files and output in build dir. consumers should be able to just load `stylesheets/defaults/index.css`
- stylus files elsewhere in `lib` do what they already do (get translated into JS modules exporting a big CSS string)

Consumers will need to change the file location to pull in `stylesheets/common.styl` (now `stylesheets/mixins/common.styl`)


jbwyme:
Nice I like this organization a lot better. Thanks for the refactoring of the build.js as well! LGTM.

PR #90, opened 2016-12-05
bookmarks widget

jbwyme:
http://cometapp.us/c/JQw48Uyj


examples/style-guide/index.js
```diff
@@ -7,22 +7,22 @@ import { Component } from 'panel';
...
 
-
 document.registerElement('style-guide', class extends Component {
   get config() {
     return {
       defaultState: {
         blueToggleValue: 'option1',
-        menuOpen: false,
-
+        bookmarks: bookmarks.bookmarks,
```
tdumitrescu: yo dawg i heard you like bookmarks...

examples/style-guide/index.js
```diff
@@ -7,22 +7,22 @@ import { Component } from 'panel';
...
         COLORS,
         SVG_ICONS,
         inputGroupSaving: false,
         open: {},
```
tdumitrescu: It's not very obvious what this map is supposed to have in it; how about making the default state:
```js
{
  open: {
    'bookmarks-widget': false,
    'menu': false,
  },
}
```

examples/style-guide/index.js
```diff
@@ -49,8 +49,32 @@ document.registerElement('style-guide', class extends Component {
...
+          this.update();
+        },
+        toggleBookmarksMenu: () => {
+          this.state.open['bookmarks-widget'] = !this.state.open['bookmarks-widget'];
+          this.update();
+        },
+        handleBookmarksMenuSubmit: e => {
+          switch (e.detail.action) {
+            case 'select':
+              console.log('Selected:' + e.detail.value.id);
```
tdumitrescu: space after :

lib/components/drop-menu/index.js
```diff
@@ -73,9 +73,11 @@ registerMPComponent('mp-drop-menu', class extends Component {
       switch(this.state.visibility) {
         case VISIBILITY_OPENING:
           this.update({visibility: VISIBILITY_OPEN});
+          this.dispatchEvent(new CustomEvent('change', {detail: {state: this.state.visibility}}));
```
tdumitrescu: seems like it would be more intuitive as:
```js
{detail: {visibility: this.state.visibility}}
```

lib/components/tooltip/index.js
```diff
@@ -1,6 +1,6 @@
 import { Component } from 'panel';
 
-import { registerMPComponent } from  '../registration.js';
+import { registerMPComponent } from  '../../registration.js';
```
tdumitrescu: If this module is moving down into `lib`, let's give it a more obvious name, like `component-registry.js`

lib/stylesheets/mixins/icons.styl
```diff
@@ -24,3 +24,28 @@ svg-icon-style(base-color=null, hover-color=null, speed=0.2s, size=null, pos-lef
         svg, path
           color hover-color
           fill hover-color
+
+icon-input(icon-color=grey-500, icon-color-focus=grey-700)
```
tdumitrescu: sweeet

examples/style-guide/index.js
jbwyme: i love bookmarks

examples/style-guide/index.js
jbwyme: +1

examples/style-guide/index.js
jbwyme: better yet `, `

lib/components/drop-menu/index.js
jbwyme: striving for consistency with `mp-modal` -- don't disagree with the naming change but I think it should be all or nothing

lib/components/tooltip/index.js
jbwyme: +1

lib/widgets/bookmarks-widget/index.styl
```diff
@@ -0,0 +1,176 @@
...
+        height 22px
+        margin-right 4px
+        svg-icon-style(grey-700)
+
+      &:hover
+        background-color rgba(255, 255, 255, 0.35)
+        color grey-900
+
+        .btn-icon
+          svg-icon-style(grey-900, speed: 0s)
```
tdumitrescu: Too bad that you have to define a separate hover style this way. The intended usage is that `svg-icon-style` gets applied to whichever element represents the entire hover area that activates the icon (so e.g. in the style guide, hovering the text next to the icon also activates the hover color); you shouldn't normally need to call the mixin twice.

lib/widgets/bookmarks-widget/index.styl
```diff
@@ -0,0 +1,176 @@
...
+            color grey-300
+            font-weight weight-regular
+
+        .delete-button
+          display none
+          flex 0
+          height 22px
+          margin-right 6px
+
+          svg-icon-style(grey-200, mp-red, speed: 0.1s, size: 22px)
```
tdumitrescu: isn't 22px already the default?

lib/widgets/bookmarks-widget/index.jade
```diff
@@ -0,0 +1,69 @@
...
+  )
+    .mp-bm-menu-list-item-content
+      .mp-bm-menu-name(attributes={title: bookmark.name})
+        +highlight(bookmark.name, filterText, 'mp-bm-menu-highlight')
+      .mp-bm-menu-creator #{bookmark.user}
+    .mp-bm-menu-delete-button(onclick=e => $helpers.confirmDeleteBookmark(e, bookmark))
+      svg-icon(attributes={icon: 'trashcan'})
+
+mixin highlight(target, search, highlightClass)
+  - const searchRegex = new RegExp(`(${search})`, 'gi');
```
jbwyme: I think having it consistent would be great

lib/widgets/bookmarks-widget/index.js
```diff
@@ -0,0 +1,261 @@
...
+            this._selectBookmark(this.state.activeBookmark);
+            break;
+          case 38: // up arrow
+            e.preventDefault();
+            activeBookmark = this.state.filteredBookmarks[activeIdx === 0 ? bookmarks.length - 1 : activeIdx - 1];
+            this._setActiveBookmark(activeBookmark);
+            break;
+          case 40: // down arrow
+            e.preventDefault();
+            activeBookmark = this.state.filteredBookmarks[activeIdx === bookmarks.length - 1 ? 0 : activeIdx + 1];
```
jbwyme: We'll have to agree to disagree I guess and defer to product. I love the loop behavior. However, the existing bookmarks widget doesn't have it so that seems like a good tie breaker in this case.

lib/widgets/bookmarks-widget/index.js
```diff
@@ -0,0 +1,261 @@
...
+  }
+
+  _filterBookmarks() {
+    const filteredBookmarks = this.state.bookmarks
+      .filter(b => b.name.toLowerCase().includes(this.state.filterText.toLowerCase()))
+      .sort((a, b) => this._sortBookmarks(a, b));
+    this.update({filteredBookmarks});
+  }
+
+  _sortBookmarks(a, b) {
```
jbwyme: but then we don't have access to `this.state`

tdumitrescu: Ah sorry, missed that

lib/widgets/bookmarks-widget/index.styl
jbwyme: good call, I can fix that

PR #91, opened 2016-12-05
Add mp-tag component

zachwe:


examples/style-guide/index.jade
```diff
@@ -191,3 +191,10 @@
       a
         | Hover me
         mp-tooltip This is my tooltip
+  .section
```
tdumitrescu: Could you get these last two components into alphabetical order in the style guide (`mp-drop-menu` and `mp-tag`)? Gets more important as the component set grows, which it is doing rapidly right now.

lib/components/tag/index.jade
```diff
@@ -0,0 +1,8 @@
+div(class='mp-tag tag')
+  - const value = $component.getAttribute('value')
+  span= value
```
tdumitrescu: Is the idea that tag values should always be plaintext? If we foresee needing more flexibility, it probably makes more sense to use shadow dom `content` as the main API from the start. @jbwyme do you have more context around this component?

Even if not, I'd recommend removing the superfluous variable:
```jade
span= $component.getAttribute('value')
```

lib/components/tag/index.jade
```diff
@@ -0,0 +1,8 @@
+div(class='mp-tag tag')
```
tdumitrescu:
```jade
.map-tag.tag
```

lib/components/tag/index.jade
```diff
@@ -0,0 +1,8 @@
...
+  - const value = $component.getAttribute('value')
+  span= value
+  svg-icon(
+    attributes={icon: "x"}
+    class='x'
```
tdumitrescu:
```jade
svg-icon.x(
```
(though I would highly recommend avoiding one-letter classnames)

lib/components/tag/index.js
```diff
@@ -0,0 +1,21 @@
...
+import template from './index.jade';
+
+registerMPComponent('mp-tag', class extends Component {
+  get config() {
+    return {
+      css,
+      template,
+      useShadowDom: true,
+      helpers: {
+        removeTag: () => {
```
tdumitrescu: It seems to me like this helper is misnamed, since it looks like it's up to the app to "remove" this element. Maybe something like `clickClose` is more accurate.

lib/components/tag/index.js
```diff
@@ -0,0 +1,21 @@
...
+          this.dispatchEvent(new Event('change'));
```
tdumitrescu: Since you're dispatching a `change` event here, it would be nice to make it a `CustomEvent` and specify in its `detail` that it's supposed to `close`.

Soon we should be able to grab real arbitrary custom events, at which point this could change to just `new CustomEvent('close')`

lib/components/tag/index.styl
```diff
@@ -0,0 +1,21 @@
+@import '../../stylesheets/defaults'
+
+.mp-tag.tag
+  align-items center
+  background-color grey-900
+  border 1 px solid black
+  border-radius 42px
+  display inline-flex
+  font-family HelveticaNeue
```
tdumitrescu: don't we have standardized font settings in a mixin now?

lib/components/tag/index.styl
```diff
@@ -0,0 +1,21 @@
...
+  font-size 11px
+  font-weight bold
+  height 24px
+  justify-content center
+  svg-icon-style(base-color:grey-500, size:15px)
```
tdumitrescu: super nitpicky space after :

lib/components/tag/index.jade
jbwyme: is that class even necessary?

lib/components/tag/index.jade
```diff
@@ -0,0 +1,8 @@
+.mp-tag.tag
+  - const value = $component.getAttribute('value')
+  span= value
```
jbwyme: no reason to assign to variable and then only use it once, just inline it

lib/components/tag/index.js
```diff
@@ -0,0 +1,21 @@
+import { Component } from 'panel';
+
+import { registerMPComponent } from  '../../components/registration.js';
+
+import css from './index.styl';
+import template from './index.jade';
+
+registerMPComponent('mp-tag', class extends Component {
```
jbwyme: do you think this should be a stand-alone component? The only behavior here is dispatching an event on click, which already happens natively.

jbwyme: If the only purpose is for styling, just make a mixin

zachwe: Yeah this is mainly just a set of shared styles, I'm definitely open to it being just a css mixin or whatever the best way is to make styles consistent across our repos.

jbwyme: to elaborate a bit more, I'm assuming this is going to be a building block for the tag autocomplete, correct? If so it might just be worth having the tags styled with a mixin and make a `widget` for the tag autocomplete

lib/components/tag/index.styl
```diff
@@ -0,0 +1,20 @@
+@import '../../stylesheets/defaults'
+
+.mp-tag.tag
+  align-items center
+  background-color grey-900
+  border 1 px solid black
```
jbwyme: `1px`

lib/components/tag/index.styl
```diff
@@ -0,0 +1,20 @@
...
+  border-radius 42px
+  display inline-flex
+  font-size 11px
```
jbwyme: we shouldn't be using font-size 11px anywhere should we @jayjo?

lib/components/tag/index.styl
```diff
@@ -0,0 +1,20 @@
...
+  font-weight bold
```
jbwyme: use the variable here: `font-weight weight-bold`

lib/components/tag/index.styl
```diff
@@ -0,0 +1,20 @@
...
+  height 24px
```
jbwyme: i'd suggest using the `+host-styles()` mixin for setting height so that it's possible for the caller to have more control over it

zachwe:
Closing this, just handling these styles as mixin as recommended

PR #92, opened 2016-12-05
fixed all of the issues that were brought up in the  branch

jayjo:


examples/style-guide-new/index.jade
```diff
@@ -0,0 +1,177 @@
...
+        .divider
+          sup 1.3
+        label.title Color Library
+        p.desc.max Lorem ipsum dolor sit amet, consectetur adipisicing elit. Vero, dolores a ratione labore recusandae accusamus maiores deleniti quis doloremque earum.
+
+        ul.color-library.reds
+          strong.call-out.floating-label mp-
+            label red
+          li.red-700
+            input.med(value='#8260D1')
```
jbwyme: we shouldn't be referring to the hex value at all in code: should be something like `colors['red-700']`

jayjo: Dammit Eric.

PR #94, opened 2016-12-08
Add dash widget

zachwe:
Basic version of an "add to dash" widget that allows a user to select from existing plaintext "tags" and create new tags.


jbwyme:
I'd let the client handle the data transfer and have the widget be strictly UI focused.

lib/widgets/tag-selector/index.js
```diff
@@ -0,0 +1,98 @@
...
+      }
+    } else if (attr === 'load_state') {
+      const loadState = this.getAttribute('load_state');
+      if (LOAD_STATES.indexOf(loadState) !== -1) {
+        this.update({loadState: this.getAttribute('load_state')});
+      }
+    }
+  }
+
+  filterOptions(value) {
```
jbwyme: for internal methods, I've been pseudo-privatizing them by prefixing with `_`: `_filterOptions`. That way they are marked as private and therefore not supported as a public API.

lib/widgets/tag-selector/index.js
```diff
@@ -0,0 +1,98 @@
...
+  }
+
+  removeTag(tagName) {
+    this.setAttribute('load_state', LOADING_TAG);
+    this.dispatchEvent(new CustomEvent('change', {detail: {tagName, action: 'removeTag'}}));
+    requestAnimationFrame(() => this.el.querySelector('.search-input').focus());
+  }
+
+  addTag(tagName) {
+    if (this.state.currentTags.indexOf(tagName) !== -1) {
```
jbwyme: how about `if(this.state.currentTags.includes(tagName))`

lib/widgets/tag-selector/index.styl
```diff
@@ -0,0 +1,130 @@
+@import '../../stylesheets/defaults'
+
+tag-styles()
```
jbwyme: any reason for this being a mixin? it looks like it's only used once

lib/widgets/tag-selector/index.js
```diff
@@ -0,0 +1,98 @@
...
+      if (LOAD_STATES.indexOf(loadState) !== -1) {
+        this.update({loadState: this.getAttribute('load_state')});
+      }
+    }
+  }
+
+  filterOptions(value) {
+    return this.state.options.filter((m) => {
+      return value === '' || m.indexOf(value) !== -1;
+    }).filter((m) => this.state.currentTags.indexOf(m) === -1);
```
jbwyme: I think this could be expressed more simply:

`return this.state.options.filter(option => !this.state.currentTags.includes(option) && option.includes(value));` 

Additionally, we probably want to do a case-insensitive search.

lib/widgets/tag-selector/index.styl
```diff
@@ -0,0 +1,130 @@
...
+        margin 6px 12px 0 0
+        &.loaded
+          svg, path
+            color green-300
+            fill green-300
+        &.loaded-error
+          svg, path
+            color mp-red
+            fill mp-red
+      .arrow-return-icon
```
jbwyme: let's add a few new lines between selectors to break up the wall-o-text

lib/widgets/tag-selector/index.js
```diff
@@ -0,0 +1,98 @@
...
+  }
+
+  attributeChangedCallback(attr) {
+    super.attributeChangedCallback(...arguments);
+    if (attr === 'open' && this.initialized) {
+      const open = this.getAttribute('open');
+      this.update({open});
+      if (open) {
+        setTimeout(() => {
+          this.update({loadState: IDLE});
```
jbwyme: what's this about?

lib/widgets/tag-selector/index.js
```diff
@@ -0,0 +1,98 @@
...
+const LOADED_TAG = 'loaded_tag';
+const ERROR = 'error';
+const LOAD_STATES = [LOADING_WIDGET, IDLE, LOADING_TAG, LOADED_TAG, ERROR];
+
+registerMPElement('mp-tag-selector', class extends Component {
+  get config() {
+    return {
+      defaultState: {
+        currentTags: [],
+        options: [],
```
jbwyme: what are your thoughts on renaming these two properties: `selectedTags`, `allTags`? It might make some of the code below read slightly better.

lib/widgets/tag-selector/index.js
```diff
@@ -0,0 +1,98 @@
...
+        value: '',
+        open: true,
+        shouldFocus: true,
```
jbwyme: is this used?

lib/widgets/tag-selector/index.js
```diff
@@ -0,0 +1,98 @@
...
+const ERROR = 'error';
+const LOAD_STATES = [LOADING_WIDGET, IDLE, LOADING_TAG, LOADED_TAG, ERROR];
+
+registerMPElement('mp-tag-selector', class extends Component {
+  get config() {
+    return {
+      defaultState: {
+        currentTags: [],
+        options: [],
+        value: '',
```
jbwyme: `value` is a bit ambiguous, what about "tagText" or something like that?

lib/widgets/tag-selector/index.jade
```diff
@@ -0,0 +1,44 @@
+.mp-tag-selector-container
+  mp-drop-menu(attributes={open: open})
+    if loadState == 'loading_widget'
+      .loader
+        mp-loader
+    else
+      content(select=".slot-header")
```
jbwyme: let's just drop the `select` here. I know we talked about having multiple slots but we don't currently have a design that requires them.

lib/widgets/tag-selector/index.jade
```diff
@@ -0,0 +1,44 @@
+.mp-tag-selector-container
+  mp-drop-menu(attributes={open: open})
```
jbwyme: Assuming we get to place where `open` isn't used internally, can we do `{open: $component.isAttributeEnabled('open')}` and drop `open` from state?

lib/widgets/tag-selector/index.js
```diff
@@ -0,0 +1,98 @@
...
+      helpers: {
+        shouldShowFooter: () => this.state.value && !this.filterOptions(this.state.value).length,
+        getSearchMatches: () => this.filterOptions(this.state.value),
+        handleSearchInput: (e) => {
+          this.update({value: e.target.value});
+        },
+        handleTagCloseClick: (tagName) => {
+          this.removeTag(tagName);
+        },
+        handleKeyActions: e => {
```
jbwyme: do we intend on handling the autocomplete options like we do in other places (e.g. the bookmarks-widget that's available in the style-guide now)? Up/down keypress to highlight options, highlighting the characters that match during search, etc

lib/widgets/tag-selector/index.js
```diff
@@ -0,0 +1,98 @@
...
+      useShadowDom: true,
+    };
+  }
+
+  attachedCallback() {
+    super.attachedCallback(...arguments);
+    const currentTags = JSON.parse(this.getAttribute('currentTags'));
+    const options = JSON.parse(this.getAttribute('allTags'));
+    const open = this.getAttribute('open');
+    this.update({currentTags, options, open});
```
jbwyme: nit: let's just do these all inline instead of assigning to var first.

lib/widgets/tag-selector/index.styl
```diff
@@ -0,0 +1,130 @@
...
+      padding 0 8px
+      vertical-align middle
+    .close-icon
+      cursor pointer
+      padding-right 5px
+
+.mp-tag-selector-container
+  border-radius 5px
+  box-shadow: 0 10px 15px 0 rgba(0, 0, 0, 0.1), 0 15px 30px 0 rgba(0, 0, 0, 0.2);
+  width 378px
```
jbwyme: let's make this 100% and let the width of the `host` specify the width. You can use `+host-styles` to give it a default.

lib/widgets/tag-selector/index.styl
```diff
@@ -0,0 +1,130 @@
...
+
+  .loader
+    height 72px
+    width 378px
+    display flex
+    mp-loader::shadow
```
jbwyme: you said don't review the loader stuff but just in case: 

1. Shadow DOM v1 removes the `::shadow` selector
2. We shouldn't need to reach into these styles anyways

lib/widgets/tag-selector/index.styl
```diff
@@ -0,0 +1,130 @@
...
+      margin auto
+      padding 0 8px
+      vertical-align middle
+    .close-icon
+      cursor pointer
+      padding-right 5px
+
+.mp-tag-selector-container
+  border-radius 5px
+  box-shadow: 0 10px 15px 0 rgba(0, 0, 0, 0.1), 0 15px 30px 0 rgba(0, 0, 0, 0.2);
```
jbwyme: can you add `box-sizing border-box` to the container rules? I'm going for consistency on all components and border-box is so much better.

lib/widgets/tag-selector/index.styl
```diff
@@ -0,0 +1,130 @@
...
+    .option
+      color grey-500
+      font-size-default()
+      font-style normal
+      font-weight weight-bold
+      height 18px
+      line-height 1.5
+      padding 8px 0px 10px 9px;
+      &:hover
+        background-color #eff3f5
```
jbwyme: we shouldn't be using any hex values here: `grey-100`

lib/widgets/tag-selector/index.styl
```diff
@@ -0,0 +1,130 @@
...
+        color mp-blue
+        cursor pointer
+  .footer
+    cursor pointer
+    font-size-medium()
+    padding 14px 96px
+    &:hover
+      background-color #eff3f5
```
jbwyme: `grey-100`

lib/widgets/tag-selector/index.js
```diff
@@ -0,0 +1,98 @@
...
+
+import css from './index.styl';
+import template from './index.jade';
+
+const LOADING_WIDGET = 'loading_widget';
+const IDLE = 'idle';
+const LOADING_TAG = 'loading_tag';
+const LOADED_TAG = 'loaded_tag';
+const ERROR = 'error';
+const LOAD_STATES = [LOADING_WIDGET, IDLE, LOADING_TAG, LOADED_TAG, ERROR];
```
zachwe: I'm basing off the designs—loading widget (zpl.io/2b9JLG), idle (zpl.io/ZXzHA5), loading tag i.e. adding a tag to a report (zpl.io/ZLwLwF), tag loaded success (zpl.io/2wTqjb) and tag load error (the last thing here zpl.io/1jA6bE).

some of these could definitely be parsed out into their own variables but I thought a single state variable would be easier to keep track of

lib/widgets/tag-selector/index.js
zachwe: Yup good catch, I missed some of those aspects in the design

lib/widgets/tag-selector/index.styl
zachwe: Yeah it doesn't need to be a mixin as-is. I was initially going to break it into a common mixin location since the dashboard app uses the same tag styles

lib/widgets/tag-selector/index.jade
```diff
@@ -0,0 +1,44 @@
...
+            type="text"
+          )
+          .border
+          .arrow-return-icon
+            svg-icon(attributes={icon: "arrow-return"})
+      .list-options-wrapper
+        for option in $helpers.getSearchMatches()
+          .option(onclick=() => $helpers.addTag(option))
+            | #{option}
+      if $helpers.shouldShowFooter()
```
zachwe: Seems reasonable

lib/widgets/tag-selector/index.js
zachwe: yeah I'll get this out of here, was here for simulating like what an http call would look like

zachwe: If we had to do one on widget open as in the designs (I don't think we'll have to at least in the use of this widget on the reports page)

zachwe:
I fixed/responded to all your comments, made the style guide example more functional, and implemented text search highlighting (I copied in your `highlight` jade mixin from bookmarks)—that's probably something we want to put in a common location now or later?

Can you take another look?

tdumitrescu:
Following up on the bookmarks PR, do we want to pull out the trickier match-finding code from IRB for these two widgets?

examples/style-guide/index.js
```diff
@@ -28,9 +28,16 @@ document.registerElement('style-guide', class extends Component {
           menu: false,
           modal: false,
           popup: false,
+          tagSelector: false,
         },
         savingBookmark: false,
         selectedBookmarkId: null,
+        showTagSelector: false,
```
jbwyme: is this used?

examples/style-guide/index.js
```diff
@@ -91,6 +98,39 @@ document.registerElement('style-guide', class extends Component {
...
         },
+        showTagSelector: () => {
+          this.update({showTagSelector: !this.state.showTagSelector})
+        },
+        handleTagSelectorChange: (e) => {
+          const tagName = e.detail.tagName;
+          if (e.detail.action === 'addTag') {
+            this.update({tagSelectorLoadState: 'loading_tag'});
+            setTimeout(() => {
+              let allTags = this.state.tagSelectorData.allTags;
```
jbwyme: what if we just make `allTags` and `selectedTags` be sets instead of arrays so you don't even have to check?

examples/style-guide/index.js
```diff
@@ -91,6 +98,39 @@ document.registerElement('style-guide', class extends Component {
...
+              if (!allTags.includes(tagName)) {
+                allTags = allTags.concat([tagName]);
```
jbwyme: if we _were_ to keep this an array, why not use push instead of concat?

examples/style-guide/index.js
```diff
@@ -91,6 +98,39 @@ document.registerElement('style-guide', class extends Component {
...
+                },
+              });
+            }, 700);
+          } else if (e.detail.action === 'removeTag') {
+            this.update({tagSelectorLoadState: 'loading_tag'});
+            setTimeout(() => {
+              this.update({
+                tagSelectorLoadState: 'loaded_tag',
+                tagSelectorData: {
+                  selectedTags: this.state.tagSelectorData.selectedTags.filter((t) => t !== tagName),
```
jbwyme: if we make these sets, you can just use delete: https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Set/delete

examples/style-guide/index.js
```diff
@@ -91,6 +98,39 @@ document.registerElement('style-guide', class extends Component {
...
+                  allTags: this.state.tagSelectorData.allTags,
```
jbwyme: ? what dis fo?

lib/stylesheets/mixins/icons.styl
```diff
@@ -49,3 +51,24 @@ icon-input(icon-color=grey-500, icon-color-focus=grey-700)
     left 8px
 
     svg-icon-style(icon-color)
+
+// Much of these styles taken from https://github.com/lukehaas/css-loaders (MIT license)
+loader-styles(size=10px, time=1.1s, spinner-color=mp-blue, background-color=white, spinner-width=3px)
```
jbwyme: we don't need this anymore right?

lib/widgets/tag-selector/index.jade
```diff
@@ -0,0 +1,54 @@
+mixin highlight(target, search, highlightClass)
```
jbwyme: we should put this mixin in some common file and use it in this and bookmarks-widget, maybe `lib/widgets/jade/mixins.jade` ? Open to debate on the folder structure.

lib/widgets/tag-selector/index.jade
```diff
@@ -0,0 +1,54 @@
...
+            svg-icon(attributes={icon: "tag"} class="tag-icon")
+
+          if !$component.isAttributeEnabled('read-only')
+            for tag in selectedTags
+              .tag
+                span= tag
+                svg-icon(
+                  attributes={icon: "x"}
+                  class="close-icon"
+                  onclick=() => $helpers.handleTagCloseClick(tag)
```
jbwyme: i'd just call this `removeTag` instead of `handleTagCloseClick`, more descriptive of the action at the call site.

lib/widgets/tag-selector/index.jade
```diff
@@ -0,0 +1,54 @@
...
+                )
+          input(
+            class="search-input"
+            oninput=$helpers.handleSearchInput
```
jbwyme: `s/handleSearchInput/setInputText/`, similar to the above, if we know the action we're going to take it's best to be descriptive at the call site

lib/widgets/tag-selector/index.jade
```diff
@@ -0,0 +1,54 @@
...
+            onkeydown=$helpers.handleKeyActions
+            type="text"
+          )
+          .border(class= loadState == 'error' ? 'loaded-error' : '')
```
jbwyme: `.border(class={'loaded-error': loadState === 'error'})`

lib/widgets/tag-selector/index.jade
```diff
@@ -0,0 +1,54 @@
...
+          .arrow-return-icon
+            svg-icon(attributes={icon: "arrow-return"})
+      .list-options-wrapper
+        for option, i in $helpers.getSearchMatches()
+          .option(
+            onclick=() => $helpers.addTag(option)
+            class={'active': i == 0 && inputText !== ''}
```
jbwyme: `===`

lib/widgets/tag-selector/index.js
```diff
@@ -0,0 +1,91 @@
...
+    this.update({
+      selectedTags: JSON.parse(this.getAttribute('selected-tags')),
+      allTags: JSON.parse(this.getAttribute('all-tags')),
+    });
+  }
+
+  attributeChangedCallback(attr) {
+    super.attributeChangedCallback(...arguments);
+    if (attr === 'open') {
+      this.update();
```
jbwyme: don't need to do this as it already happens for you: https://github.com/mixpanel/panel/blob/master/lib/component.js#L261

lib/widgets/tag-selector/index.js
```diff
@@ -0,0 +1,91 @@
...
+      },
+      useShadowDom: true,
+    };
+  }
+
+  attachedCallback() {
+    super.attachedCallback(...arguments);
+    this.update({
+      selectedTags: JSON.parse(this.getAttribute('selected-tags')),
+      allTags: JSON.parse(this.getAttribute('all-tags')),
```
jbwyme: what if a load-state is set before it's attached? doesn't look like we handle it.

lib/widgets/tag-selector/index.js
```diff
@@ -0,0 +1,91 @@
...
+    });
+  }
+
+  attributeChangedCallback(attr) {
+    super.attributeChangedCallback(...arguments);
+    if (attr === 'open') {
```
jbwyme: make this a switch?

lib/widgets/tag-selector/index.js
```diff
@@ -0,0 +1,91 @@
...
+      }
+    } else if (attr === 'selected-tags') {
+      this.update({selectedTags: JSON.parse(this.getAttribute('selected-tags'))});
+    } else if (attr === 'all-tags') {
+      this.update({allTags: JSON.parse(this.getAttribute('all-tags'))});
+    }
+  }
+
+  _filterOptions(inputText) {
+    return this.state.allTags.filter(tag => !this.state.selectedTags.includes(tag) && tag.toLowerCase().includes(inputText.toLowerCase()));
```
jbwyme: multiline

lib/widgets/tag-selector/index.styl
jbwyme: yeah let's get rid of the mixin

lib/widgets/tag-selector/index.styl
```diff
@@ -0,0 +1,162 @@
...
+    border-top 1px solid grey-200
+    padding 0 20px;
+    .option
+      color grey-500
+      font-size-default()
+      font-style normal
+      font-weight weight-medium
+      height 18px
+      line-height 1.5
+      padding 8px 0px 10px 9px;
```
jbwyme: remove `;`

lib/widgets/tag-selector/index.js
```diff
@@ -0,0 +1,91 @@
...
+        getSearchMatches: () => this._filterOptions(this.state.inputText),
+        handleSearchInput: (e) => {
+          this.update({inputText: e.target.value});
+        },
+        handleTagCloseClick: (tagName) => {
+          this._removeTag(tagName);
+        },
+        handleKeyActions: e => {
+          if (e.keyCode === 13) { // enter
+            this._addTag(this.state.inputText);
```
jbwyme: shouldn't this select the highlighted item if there is one? http://cometapp.us/c/eTzfBUuv

lib/widgets/tag-selector/index.js
```diff
@@ -0,0 +1,91 @@
...
+          } else if (e.keyCode === 8 && this.state.selectedTags.length) { // backspace
+            if (this.state.inputText.length === 0) {
+              this._removeTag(this.state.selectedTags[this.state.selectedTags.length - 1]);
+            }
+          }
```
jbwyme: what about up/down arrow keys for navigating the options?

lib/widgets/tag-selector/index.js
jbwyme: also the footer should reset after _addTag is called shouldn't it? http://cometapp.us/c/REcJrxCq

lib/widgets/tag-selector/index.jade
```diff
@@ -0,0 +1,54 @@
...
+      .loader
+        mp-loader
+    else
+      content
+      .top-input-wrapper
+        .input-wrapper 
+          if loadState == 'loading_tag'
+            mp-loader
+          else if loadState == 'loaded_tag'
+            svg-icon(attributes={icon: "tag"} class="tag-icon loaded")
```
jbwyme: this behavior seems useless to me since it stays green once you add one tag. It'd make sense if it showed green and then transitioned back to grey.

lib/widgets/tag-selector/index.js
jbwyme: yeah i dont know, im still not sold on this yet but i don't have the right solution in mind so we can revisit once we get closer to the end

jbwyme:
@tdumitrescu yes we *do* want to do that. I was mulling it over last night about how to make it generic and at what level of abstraction it would exist. do you have a proposal for how we might implement the behavior generically?

tdumitrescu:
I'm not sure of what the requirements are for autocomplete, and whether they're different from other searches. Here's the jsdoc of the matching function from IRB for a single string + single set of search term input, maybe you can judge whether it would be useful:
```js
/**
 * Split a string into match and non-match substrings based on finding
 * all whitespace-separated terms in a given filter string (for UI search
 * bars).
 * @param {string} str - string to search
 * @param {string} filterStr - string with filter/search terms
 * @returns {Array} list of alternating matching and non-matching substrings
 * in order of the original string; even = match, odd = no match
 * @example
 * stringFilterMatches('my example string', 'ex my');
 * // ['my', ' ', 'ex', 'ample string']
 */
```
Basically in a UI, each item in the returned array becomes a `<span>` with alternating match/non-match classes for styling. Here are the test cases:
```js
describe('stringFilterMatches', function() {
  it('matches at the beginning of a string', function() {
    expect(stringFilterMatches('abcdefg', 'abc')).to.eql(['abc', 'defg']);
  });

  it('matches in the middle/end of a string', function() {
    expect(stringFilterMatches('abcdefg', 'def')).to.eql(['', 'abc', 'def', 'g']);
    expect(stringFilterMatches('abcdefg', 'efg')).to.eql(['', 'abcd', 'efg', '']);
  });

  it('returns null for non-matches', function() {
    expect(stringFilterMatches('abcdefg', 'ac')).to.eql(null);
  });

  it('is case-insensitive', function() {
    expect(stringFilterMatches('abcdefg', 'dEf')).to.eql(['', 'abc', 'def', 'g']);
    expect(stringFilterMatches('abCDEfg', 'dEF')).to.eql(['', 'abC', 'DEf', 'g']);
  });

  it('ignores whitespace padding', function() {
    expect(stringFilterMatches('abcdefg', '   def')).to.eql(['', 'abc', 'def', 'g']);
    expect(stringFilterMatches('abcdefg', 'abc   ')).to.eql(['abc', 'defg']);
    expect(stringFilterMatches('abcdefg', '  abc ')).to.eql(['abc', 'defg']);
  });

  it('matches when no filter string is passed', function() {
    expect(stringFilterMatches('abcdefg', '')).to.eql(['', 'abcdefg']);
    expect(stringFilterMatches('abcdefg', null)).to.eql(['', 'abcdefg']);
  });

  it('matches when all space-separated terms match', function() {
    expect(stringFilterMatches('abcdefg', 'abc efg')).to.eql(['abc', 'd', 'efg', '']);
    expect(stringFilterMatches('abcdefg', 'abc   efg')).to.eql(['abc', 'd', 'efg', '']);
    expect(stringFilterMatches('abcdefg', '  abc   efg   ')).to.eql(['abc', 'd', 'efg', '']);
  });

  it('matches out of order', function() {
    expect(stringFilterMatches('abcdefg', 'efg abc')).to.eql(['abc', 'd', 'efg', '']);
  });

  it('merges contiguous matches', function() {
    expect(stringFilterMatches('abcdefg', 'efg bcd')).to.eql(['', 'a', 'bcdefg', '']);
  });

  it('does not match when one or more space-separated terms do not match', function() {
    expect(stringFilterMatches('abcdefg', 'abc xxx')).to.eql(null);
  });
});
```

jbwyme:
haha nice that's basically what we did with this mixin: https://github.com/mixpanel/mixpanel-common/pull/94/commits/7e2ed7fdbaf38a1d92b5b9c4891a5a8a315c8d83#diff-3f260d6e12ca0d660863f2567684da01R1

tdumitrescu:
Yup, we got a jade mixin for it in irb too.

jbwyme:
well start contributing your shit to mixpanel-common homie.

tdumitrescu:
So far it hasn't been needed elsewhere. I offered when I saw you doing filtering in the BMW!

lib/widgets/tag-selector/index.jade
zachwe: Yeah this is in the designs but idk what the optimal behavior is. Maybe it should become grey again once the user starts typing

lib/stylesheets/mixins/icons.styl
zachwe: yeah it's because this was rebased off the old loader commit, i'll rebase master

lib/widgets/tag-selector/index.js
zachwe: added below

lib/widgets/tag-selector/index.js
zachwe: Yup, agree on both of these

lib/widgets/tag-selector/index.js
```diff
@@ -0,0 +1,91 @@
...
+  }
+
+  _removeTag(tagName) {
+    this.setAttribute('load_state', LOADING_TAG);
+    this.dispatchEvent(new CustomEvent('change', {detail: {tagName, action: 'removeTag'}}));
+    requestAnimationFrame(() => this.el.querySelector('.search-input').focus());
+  }
+
+  _addTag(tagName) {
+    if (this.state.selectedTags.includes(tagName)) {
```
jbwyme: let's just make this one block
```
if (!this.state.selectedTags.incudes(tagName)) {
  this.setAttribute......
}
```

lib/widgets/tag-selector/index.styl
```diff
@@ -0,0 +1,155 @@
+@import '../../stylesheets/defaults'
+
++host-styles(mp-tag-selector)
+  width: 378px
+
++host-styles(mp-drop-menu)
+  display block
+  width 378px
```
jbwyme: how does this width and the one above relate to each other?

lib/widgets/tag-selector/index.js
```diff
@@ -37,7 +37,7 @@ registerMPElement('mp-tag-selector', class extends Component {
           this._removeTag(tagName);
         },
         handleKeyActions: e => {
-          if (e.keyCode === 13) { // enter
+          if (e.key === 'Enter') {
```
jbwyme: why'd you switch to this?

test/widgets/tag-selector.js
```diff
@@ -0,0 +1,51 @@
+/* global beforeEach, describe, it */
+import expect from 'expect.js';
+import '../../lib/widgets/tag-selector';
+
+const allTags = ['tag foo', 'tag bar', 'tag baz'];
+const selectedTags = ['tag foo'];
+
+describe('Test tag-selector widget', () => {
+  beforeEach(function(done) {
```
jbwyme: no need for `done` if there's no async code in all of these functions

test/widgets/tag-selector.js
```diff
@@ -0,0 +1,51 @@
...
+    this.widget.el.querySelectorAll('.option').forEach((el) => {
+        displayedTags.push(el.innerText);
+    });
+    expect(displayedTags).to.eql(['tag bar', 'tag baz']);
+    done();
+  });
+
+  it('fires change event to add tags', function(done) {
+    this.widget.addEventListener('change', e => {
+      console.log(Object.keys(e));
```
jbwyme: remove logs

lib/widgets/tag-selector/index.jade
```diff
@@ -0,0 +1,54 @@
+mixin highlight(target, search, highlightClass)
+  - const searchRegex = new RegExp(`(${search})`, 'gi');
+  - const tokens = target.replace(searchRegex, '<>$1<>').split('<>');
+  for token in tokens.filter(Boolean)
+    span(class={[highlightClass]: token.toLowerCase() === search.toLowerCase()})= token
+
+.mp-tag-selector-container
+  mp-drop-menu(attributes={open: $component.isAttributeEnabled('open')})
```
jbwyme: how much work would it be to get this out of here like we talked about last week? I'd prefer not to merge this with the drop-menu if we can avoid it.

lib/widgets/tag-selector/index.js
zachwe: It looks like `keyCode` is deprecated (https://developer.mozilla.org/en-US/docs/Web/API/KeyboardEvent/keyCode), and it wasn't being set in tests even if passed into the EventInit object

lib/widgets/tag-selector/index.jade
zachwe: removed in https://github.com/mixpanel/mixpanel-common/pull/94/commits/a7f02fe7a1dfff619c8661ccc57b9fb08fc63944

lib/widgets/tag-selector/index.js
jbwyme: support for it may be deprecated but it's not dropped and `key` is not fully supported by a lot of browsers yet (and not at all in Safari): https://developer.mozilla.org/en-US/docs/Web/API/KeyboardEvent/key#Browser_compatibility

test/widgets/tag-selector.js
```diff
@@ -0,0 +1,45 @@
...
+  it('displays all tags that are not selected', function() {
+    let displayedTags = [];
+    this.widget.el.querySelectorAll('.option').forEach((el) => {
+        displayedTags.push(el.innerText);
+    });
+    expect(displayedTags).to.eql(['tag bar', 'tag baz']);
+  });
+
+  it('fires change event to add tags', function() {
+    this.widget.addEventListener('change', e => {
```
jbwyme: this method *is* async so it needs the `done` behavior

test/widgets/tag-selector.js
```diff
@@ -0,0 +1,45 @@
...
+      expect(e.detail.tagName).to.equal('new tag');
+    });
+    const searchInput = this.widget.el.querySelector('.search-input');
+    searchInput.value = 'new tag';
+    searchInput.dispatchEvent(new Event('input'));
+    searchInput.dispatchEvent(new KeyboardEvent('keydown', {code: 13, key: "Enter", keyCode: 13}));
+  });
+
+  it('fires change event to remove tags', function() {
+    this.widget.addEventListener('change', e => {
```
jbwyme: same with this one

zachwe:
doh, fixed and rebased + updated a bunch of stuff to pass new linter rules

PR #95, opened 2016-12-09
Add mp-loader component

zachwe:


lib/components/loader/index.jade
```diff
@@ -0,0 +1 @@
+.mp-loader
```
tdumitrescu: OK I know I suggested that a custom element would be a good API for this, but I wasn't aware that it would involve no markup and no js. I.e. shouldn't it just be a CSS class? @jbwyme opinions?

lib/components/index.js
```diff
@@ -3,6 +3,7 @@ import './button-bar';
 import './confirm';
 import './confirm-delete';
 import './input-group';
+import './loader';
```
jbwyme: let's go with `spinner` instead of loader so it's a bit more explicit

jbwyme:
After looking through the code and thinking about some more, there are two things that I think we should do our best to expose an API for: size and color. If we use an SVG for the spinner, size becomes easy to solve for: just make it 100% height and width. Then we have to solve for color. I'm working with design right now to figure out how we want to handle color. IMO restricting the API as much as possible is the right thing to do in this case. It prevents a bunch of customization being done which a) makes it harder to modify components without breaking people's UIs and b) makes mixpanel less consistent which is a major goal of `mixpanel-common`.

lib/components/loader/index.styl
```diff
@@ -0,0 +1,21 @@
+@import '../../stylesheets/defaults'
+
+// Much of these styles taken from https://github.com/lukehaas/css-loaders (MIT license)
```
zachwe: Definitely on board with getting sign-off from design prior to merging

lib/stylesheets/mixins/icons.styl
```diff
@@ -24,3 +26,24 @@ svg-icon-style(base-color=null, hover-color=null, speed=0.2s, size=null, pos-lef
         svg, path
           color hover-color
           fill hover-color
+
+// Much of these styles taken from https://github.com/lukehaas/css-loaders (MIT license)
+loader-styles(size=10px, time=1.1s, spinner-color=mp-blue, background-color=white, spinner-width=3px)
```
zachwe: Oops, meant to remove this. If we are doing it as a component, not a mixin it's probably best to kill this and just keep it in loader/index.styl right?

zachwe:
@jbwyme updated, PTAL. The way we should handle color imo is by a finite number of options dictated by class name. However I left that out of this PR (just defaulted to mp-blue) since we don't seem to know right now what those other options are.

jbwyme:
looks great, thanks for reworking this

PR #96, opened 2016-12-13
mp-alert

rickydoar:
mp-alert modal

### DIFF
--- a/lib/components/alert/index.jade
+++ b/lib/components/alert/index.jade
@@ -1,0 +1,9 @@
+- const proxyAttrs = {'modal-type': 'popup'};
+- ['closeable', 'open', 'width'].forEach(attr => proxyAttrs[attr] = $component.getAttribute(attr));
+mp-modal(attributes=proxyAttrs)
+  .mp-alert-container
+    .mp-alert-body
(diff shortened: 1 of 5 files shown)


tdumitrescu:
this should be an `mp-modal` IMO

jbwyme:
@tdumitrescu why do you think that?

tdumitrescu:
Because it doesn't seem any more different from the other 4 or so existing modal types than they are from each other. But I don't have a lot of details, so don't feel real strongly.

lib/components/alert/index.jade
```diff
@@ -0,0 +1,9 @@
+- const proxyAttrs = {'modal-type': 'popup'};
+- ['closeable', 'open', 'width'].forEach(attr => proxyAttrs[attr] = $component.getAttribute(attr));
```
jbwyme: no need to proxy `width` anymore, I've made it so you can specify it using the css. Make sure you're up to date with the latest master and take a look at the `mp-confirm` example: https://github.com/mixpanel/mixpanel-common/blob/master/lib/components/confirm/index.styl#L8

PR #101, opened 2016-12-15
Update linter to use shared config

jbwyme:


tdumitrescu:
backtix ahoy! can we run the linter as step 1 of `npm test`?

PR #102, opened 2016-12-16
add stylint

jbwyme:


lib/stylesheets/mixins/shadowdom.styl
```diff
@@ -1,22 +1,22 @@
 content-styles()
   &
     {block}
-  content::content
-    {block}
+  content::content // @stylint ignore
+    {block} // @stylint ignore
```
tdumitrescu: What's the rule being ignored here?

jbwyme: "duplicate property or selector, consider merging [duplicates: true]"

It wants me to combine the selectors since the content is the same (which normally makes sense). In this case though, we cannot do that.

PR #104, opened 2016-12-17
add mp-box-shadow mixin

evnp:


lib/components/drop-menu/index.styl
```diff
@@ -1,4 +1,5 @@
 @import '../../stylesheets/defaults'
+@import '../../stylesheets/mixins/common.styl'
```
tdumitrescu: i think this already comes free with `stylesheets/defaults`; does it not work without the explicit `common.styl` import?

PR #106, opened 2016-12-22
added esc listener to close menu

rickydoar:
mp-drop-menu now closes when you hit escape

### DIFF
--- a/examples/style-guide/index.jade
+++ b/examples/style-guide/index.jade
@@ -108,7 +108,10 @@
     p
       button(onclick=$helpers.toggleMenu) Open Menu
       .drop-menu-container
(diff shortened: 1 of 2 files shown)


examples/style-guide/index.jade
```diff
@@ -108,7 +108,10 @@
     p
       button(onclick=$helpers.toggleMenu) Open Menu
       .drop-menu-container
-        mp-drop-menu(attributes={open: open.menu})
+        mp-drop-menu(
+        attributes={open: open.menu}
```
jbwyme: you don't like tabs or what?

rickydoar: tabs are just holding me back

PR #109, opened 2016-12-26
Add permissions restrictions to bookmarks widget

ldelbeccaro:
- take away delete button in each bookmark list item
- take away main create button

### DIFF
--- a/lib/widgets/bookmarks-widget/index.js
+++ b/lib/widgets/bookmarks-widget/index.js
@@ -16,7 +16,6 @@ registerMPElement(`mp-bookmarks-widget`, class extends Component {
         createMode: false,
         filteredBookmarks: [],
         filterText: ``,
-        hasWritePermissions: true,
         name: ``,
         open: false,
         selectedBookmark: null,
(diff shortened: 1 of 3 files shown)


lib/widgets/bookmarks-widget/index.js
```diff
@@ -147,14 +148,8 @@ registerMPElement(`mp-bookmarks-widget`, class extends Component {
...
-      case `selected-bookmark-id`: {
-        this._resetActiveBookmark();
-        break;
-      }
-      case `user-id`: {
-        this.state.userId = parseInt(this.getAttribute(`user-id`));
-        this._filterBookmarks();
-        this._resetActiveBookmark();
+      case `has_write_permissions`: {
+        this.update({hasWritePermissions: this.isAttributeEnabled(`has-write-permissions`)});
```
tdumitrescu: this doesn't work does it? `has-write-permissions` !== `has_write_permissions`

BTW if all you're doing is reflecting the component attribute directly into a state member, you _could_ instead just reference it in the template directly with `$component.isAttributeEnabled('has-write-permissions')`. Components are smart enough to re-render themselves when attributes change.

ldelbeccaro: sweet, done. thanks!

PR #110, opened 2017-01-03
migrate to snabbdom

tdumitrescu:
Using 'beta' release of Panel from https://github.com/mixpanel/panel/pull/15.

Mostly just template changes, with a little extra `class` twiddling since we won't be using string notation anymore (`class='foo-bar'` becomes `class={'foo-bar': true}`, which we already use for dynamic classes and which snabbdom handles 'natively' with its class module).

Main benefits we'll get out of this:
- support for adding/removing/delaying [class](https://github.com/tdumitrescu/snabbdom-delayed-class#delayed-classes) and [style](https://github.com/snabbdom/snabbdom#delayed-properties) attributes during the rendering cycle
- [lifecycle hooks](https://github.com/snabbdom/snabbdom#usage) to help with things like inspecting rendered element attributes during the DOM-manipulation cycle
- listener support for custom event names thanks to the [Eventlisteners module](https://github.com/snabbdom/snabbdom#eventlisteners-module)


build.js
```diff
@@ -28,6 +28,8 @@ const CONFIGS = [
           filename: filename,
           name: '_jade_template_fn',
           pretty: true,
+          vdom: 'snabbdom',
+          runtime: `var h = require('panel').h;`,
```
jbwyme: how does the runtime stuff work? just curious

lib/components/utils.js
```diff
@@ -1,5 +1,14 @@
 export const ANIMATION_END_EVENTS = [`webkitAnimationEnd`, `animationend`, `oAnimationEnd`, `MSAnimationEnd`];
 
+
+export function extendComponentClasses(extraClasses) {
```
jbwyme: FWIW I was thinking about the pattern that requires this function and I believe it's really an anti-pattern. Exposing the api to a web component via classes is weird and would prefer we use attributes moving forward. e.g: `<mp-button theme="red">` vs `<mp-button class="mp-button-red">`. Functionally it's the same thing but IMO it's a better API.

build.js
tdumitrescu: It gets stuck at the top of each compiled template. Panel for convenience exposes `h` from whatever virtual dom lib it's using, and by referencing it here we ensure that the compiled templates use the exact same lib+version (for instance, not needing to change when switching from `virtual-dom` to `snabbdom-tdumitrescu` to `snabbdom`).

lib/components/utils.js
tdumitrescu: I agree. This change was purely about converting the existing to code to work with class-object notation. Changing the components' public API should be a future step IMO.

jbwyme: agreed just wanted to point it out, maybe even worth a comment in this file

tdumitrescu: word https://github.com/mixpanel/mixpanel-common/pull/110/commits/4bde6ae45de22d8dffb8d4a0c1296aea2e194c57

package.json
```diff
@@ -1,16 +1,19 @@
...
     "build": "node build.js",
     "build-examples": "npm run build && webpack --context examples --config examples/webpack.config.js",
+    "build-test": "webpack --context test --config test/webpack.config.js",
     "watch": "npm run build && concurrently 'node build.js -w' 'webpack --context examples --config examples/webpack.config.js -w'",
     "start": "npm run build && concurrently 'node build.js -w' 'webpack-dev-server --context examples --config examples/webpack.config.js --colors --inline'",
     "lint": "eslint lib && stylint lib",
     "prepublish": "npm run build",
-    "test": "npm run lint && webpack --context test --config test/webpack.config.js; DISPLAY=:30 wct --plugin local",
+    "test": "npm run lint && npm run build-test && wct --plugin local",
+    "test-p": "wct --plugin local --persistent",
```
jbwyme: is the reason for not building here because we assume you're debugging and have already built?

tdumitrescu: basically. my own workflow is that the only time i run persistent is after seeing a non-obvious failure in terminal output. I don't feel strongly about it.

PR #111, opened 2017-01-05
`tag-selector` iteration

jbwyme:
- Cleaned up a number of UX issues
- Added the ability to select options using the up/down arrow keys
- Factored out list styles to `mp-list-item`
- Added generic function for autocomplete

There are still a few things to do:
- wire loading states back up
- potentially remove backspace behavior
- componentize tags
- fix click outside with nested shadowdom elements


lib/util/string.js
```diff
@@ -0,0 +1,55 @@
+/**
```
jbwyme: this is copied from IRB so no need to review but we should get IRB updated to use this one

test/util/string.js
```diff
@@ -0,0 +1,55 @@
+import expect from 'expect.js';
```
jbwyme: copied from IRB

lib/components/list-item/index.jade
```diff
@@ -0,0 +1,6 @@
+div(class={
+  'mp-list-item-container': true,
+  'mp-list-item-active': $component.isAttributeEnabled('active'),
+  'mp-list-item-static': $component.isAttributeEnabled('static')
```
rickydoar: Can an item be both active and static? I.e. is there a point to having them be separate attributes instead of just one attribute on vs. off?

lib/widgets/tag-selector/index.js
```diff
@@ -11,41 +12,58 @@ const LOADING_TAG = `loading_tag`;
...
             }
+            case 38: // up arrow
+              this.update({activeTagIndex: Math.max(0, this.state.activeTagIndex - 1)});
+              break;
+            case 40: // down arrow
+              this.update({activeTagIndex: Math.min(this.state.matchingTags.length - 1, this.state.activeTagIndex + 1)});
+              break;
+            default:
+              this._resetActiveTagIndex();
+              break;
```
rickydoar: so much control 👍

lib/widgets/tag-selector/index.js
```diff
@@ -54,41 +72,86 @@ registerMPElement(`mp-tag-selector`, class extends Component {
...
-    if (attr === `load-state`) {
-      const loadState = this.getAttribute(`load-state`);
-      if (LOAD_STATES.includes(loadState)) {
-        this.update({loadState});
+    if (this.initialized) {
+      switch(attr) {
+        case `load-state`: {
+          const loadState = this.getAttribute(`load-state`);
+          if (LOAD_STATES.includes(loadState)) {
+            //this.update({loadState});
```
rickydoar: is this supposed to be commented still?

lib/components/list-item/index.jade
jbwyme: Well these are unrelated: something can be both not static and not active. Static just means that hover states don't apply

lib/widgets/tag-selector/index.js
jbwyme: yeah i need to re-think how the loadState stuff is being handled

jbwyme:
Yes, I was under the impression that main-loop queued these for you and batched them. However, there seems to be something related to the multiple update calls and webcomponent.js polyfill. I just tested it using two nested requestAnimationFrame delays and it passes on FF, Safari, and IE whereas it fails with just one.

tdumitrescu:
I think the main difference you're running into is that Chrome fires `attachedCallback` immediately upon inserting the element into the DOM, whereas the others fire it on the next animation frame (probably because of the polyfill, but just possibly because of the different rendering engines). I think there are also now some subtle differences in when `attributeChangedCallback` gets called. In any case, waiting twice on `requestAnimationFrame` for your updates should keep your test deterministic across all browsers, whereas `setTimeout(..., 100)` introduces a race condition and will make the test fail if it can't get through the first frame within 100ms (could happen with e.g. heavy load on hardware running test VM + GC pause).

You're compiling this test case code into ES5 right? In that case you can avoid all the nesting by [`await`ing animation frames](https://github.com/mixpanel/panel/blob/master/test/server/component.js#L31-L41) like we do in panel server-side tests.

PR #113, opened 2017-01-06
Adding background image mixins

jalack:


tdumitrescu:
Could you make a matching IRB PR to remove the code there?

jbwyme:
It'll be mixpanel-common version 0.31.0 fyi

PR #114, opened 2017-01-07
Fix some problems with the tooltip

mduan:
1. Use mouseenter instead of mouseover since mouseover is fired when
moving into a child element.
2. Handle case when parent element has non-static position.


tdumitrescu:
Component code LGTM; any chance you could add an example to the Style Guide that exercises the 'non-static-positioned parent' code?

mduan:
Good call on adding an example with non-statically positioned parent. Caused me to find another bug which I've fixed.

tdumitrescu:
Published 0.31.1

PR #116, opened 2017-01-09
Keep dropdown open on clicks inside and close on click outside

jalack:
Working around shadow DOM retargeting


lib/components/drop-menu/index.js
```diff
@@ -94,7 +94,10 @@ registerMPElement(`mp-drop-menu`, class extends Component {
 
     // close when the user clicks outside of the menu
     this.clickOutsideListener = e => {
-      if (this.isAttributeEnabled(`open`) && wasClickedOutside(e.target, this)) {
+      // events from shadow DOM will be retargeted to appear as if they come from component
+      // use path to distinguish between events within nested components
+      const target = e.path ? e.path[0] : e.target;
+      if (this.isAttributeEnabled(`open`) && wasClickedOutside(target, this)) {
```
jbwyme: How about we change the signature of `wasClickedOutside` to be `wasClickedOutside(element, clickEvent)` so that this target mapping is abstract out? Then create a test in `test/utils/dom.js` that tests the possible scenarios (no shadowDOM, shadowDOM, nested ShadowDOM). I'd recommend changing the signature, writing the tests, confirm the nested shadowDOM test fails, and then adding the `e.path` fix.

test/widgets/bookmarks-widget.js
```diff
@@ -44,4 +44,23 @@ describe('Test boomarks-widget', function() {
     });
     this.widget.setAttribute('open', 'true');
   });
+
+
+  it('does not close when element inside widget is clicked', function(done) {
```
jbwyme: i'd just remove this test completely after writing the tests in `tests/util/dom.js`

jalack: do we not want a sanity check that just tests that close behavior works as expected? or should we just assume that with properly tested helper methods that won't be an issue?

jbwyme: I think there's something to be said about avoiding "over" testing when we technically already have full coverage. However, I'm fine with erring on the side of caution. That being said, the better place to test this would be where we actually use `wasClickedOutside` which is the `drop-menu` component.

test/util/dom.js
```diff
@@ -0,0 +1,90 @@
...
+    const childElement = document.createElement(`span`);
+    parentElement.appendChild(childElement);
+    document.body.appendChild(parentElement);
+    this.clickOutsideListener = e => {
+      expect(clickWasOutside(e, parentElement)).to.equal(false);
+    };
+
+    document.addEventListener(`click`, this.clickOutsideListener);
+    childElement.click();
+    done();
```
jbwyme: this needs to be called after line 28. Currently it's just called immediately after the click is fired and the test will always pass.

test/util/dom.js
```diff
@@ -0,0 +1,90 @@
+/* global beforeEach, afterEach, describe, it */
+
+import expect from 'expect.js';
+import { onAnimationEnd } from '../../lib/util/dom';
+import bookmarks from '../widgets/bookmark-data.json';
+
+import {
```
jbwyme: let's just inline this

test/util/dom.js
```diff
@@ -0,0 +1,90 @@
...
+  clickWasOutside,
+} from '../../lib/util/dom';
+
+describe(`clickWasOutside`, function() {
+  beforeEach(function(done) {
```
jbwyme: no need to pass `done` if the function doesn't contain async code

test/util/dom.js
```diff
@@ -0,0 +1,90 @@
...
+    done();
+  });
+
+  it(`identifies when an element within another element is clicked`, function(done) {
+    const parentElement = document.createElement(`div`);
+    const childElement = document.createElement(`span`);
+    parentElement.appendChild(childElement);
+    document.body.appendChild(parentElement);
+    this.clickOutsideListener = e => {
+      expect(clickWasOutside(e, parentElement)).to.equal(false);
```
jbwyme: `done()` needs to be called within this listener, not at the end of the method

test/util/dom.js
```diff
@@ -0,0 +1,90 @@
...
+    const childElement = document.createElement(`span`);
+    parentElement.appendChild(childElement);
+    document.body.appendChild(parentElement);
+    this.clickOutsideListener = e => {
+      expect(clickWasOutside(e, childElement)).to.equal(true);
+    };
+
+    document.addEventListener(`click`, this.clickOutsideListener);
+    parentElement.click();
+    done();
```
jbwyme: move to inside listener

test/util/dom.js
```diff
@@ -0,0 +1,90 @@
...
+
+describe(`clickWasOutside`, function() {
+  beforeEach(function(done) {
+    document.body.innerHTML = ``;
+    done();
+  });
+
+  afterEach(function(done) {
+    document.removeEventListener(`click`, this.clickOutsideListener);
+    done();
```
jbwyme: no need to pass or call done

test/util/dom.js
```diff
@@ -0,0 +1,90 @@
...
+      expect(clickWasOutside(e, childElement)).to.equal(true);
+    };
+
+    document.addEventListener(`click`, this.clickOutsideListener);
+    parentElement.click();
+    done();
+  });
+
+  it(`identifies when an element within a shadow DOM component is clicked`, function(done) {
+    const bookmarksWidget = document.createElement(`mp-bookmarks-widget`);
```
jbwyme: 1) This is relying on a dependency that isn't explicitly imported so it's not guaranteed to be anything other than an empty html element. 
2) Instead of relying on an existing widget and having to write this additional bootstrapping code, let's create and register some test components inside this test class so it's all self contained.

test/util/dom.js
```diff
@@ -0,0 +1,84 @@
+/* global beforeEach, afterEach, describe, it */
+
+import expect from 'expect.js';
+
+import {
```
jbwyme: inline this

test/util/dom.js
```diff
@@ -0,0 +1,81 @@
...
+    this.clickOutsideListener = e => {
+      expect(clickWasOutside(e, hostElement)).to.equal(false);
+      done();
+    };
+
+    document.addEventListener(`click`, this.clickOutsideListener);
+    shadowChild.click();
+  });
+
+  it(`identifies when an element within a nested shadow DOM component is clicked`, function(done) {
```
jbwyme: I think what we want to test is this:

`document` has child `shadowdom-a`
`shadowdom-a` has children `shadowdom-b`, `shadowdom-c`
`shadowdom-c` has child `shadowdom-d`


slick shadowdom-a and ensure that `clickWasOutside(e, shadowdom-c) === true`
click shadowdom-b and ensure that `clickWasOutside(e, shadowdom-c) === true`
click shadowdom-c and ensure that `clickWasOutside(e, shadowdom-c) === false`
click shadowdom-d and ensure that `clickWasOutside(e, shadowdom-c) === false`

jbwyme:
sweet, just get the build passing and i'll get it merged!

jbwyme:
fyi I squashed your commits and then rebased on tag-selector

jbwyme:
alright looks like we're down to just Safari latest failing. I'm betting this has something to do with Safari 10 having ShadowDOM v1 support: https://developer.apple.com/library/content/releasenotes/General/WhatsNewInSafari/Articles/Safari_10_0.html

jbwyme:
Yep, `e.composedPath()` exists in Safari 10 but since the polyfill wraps elements the includes never matches so we need to ensure we use `e.path` if we're polyfilled.

jalack:
awesome, thanks for the cleanup!

PR #117, opened 2017-01-10
Unique of shallow objects

evnp:


lib/util/array.js
```diff
@@ -59,8 +59,29 @@ export function sorted(arr, options={}) {
...
+}
+
+/**
+ * Return a copy of an array of shallow objects with all duplicate objects removed
+ * @param {Array} arr - array of objects to de-duplicate
+ * @param {Array} keys - attrs of each object to compare it to other objects in array
+ * An array of object keys specifies how the objects will be compared - any objects containing
+ * the same values for all specified keys will be considered duplicates and removed.
+ */
+export function uniqueOfShallowObjects(arr, keys) {
```
tdumitrescu: Is this really the right level of abstraction? I would say that if we want the util to be this specific and it's only used in IRB, then leave it in IRB right now. It's not something that someone's likely to think of looking for in a general-util tire-fire anyway.

What I might consider, however, is integrating this into `unique(arr)` with an optional second argument that specifies how to turn an array item into a key:
```js
export function unique(arr, keyFunc=x => x) {
  const encountered = new Set();
  return arr.filter(function(item) {
    const key = keyFunc(item);
    return encountered.has(key) ? false : encountered.add(key);
  });
}
```
which you could then call for your particular case with:
```js
unique([{a: 1, b: 2}, {a: 3, b: 4}, {a: 1, b: 2}], o => `${o.a}:${o.b}`);
```
or more generically
```js
unique(myObjArr, o => Object.keys(o).sort().map(k => o[k]).join(`:`));
```

FWIW I think the existing `unique` implementation should be reworked either way to use a set or object internally like this one, since the current `Array.includes()` version is easily prone to pathological running time cases.

test/util/util.js
```diff
@@ -215,3 +217,106 @@ describe('insertAtIndex()', function() {
     expect(insertAtIndex(array, array.length, '$')).to.eql(['a', 'b', 'c', 'd', 'e', '$']);
   });
 });
+
+describe('unique()', function() {
```
tdumitrescu: Thanks for the `unique` tests, much appreciated!

lib/util/array.js
evnp: If we have `unique` in mixpanel-common then a util like this also belongs in mixpanel-common IMO, since it fills a gaping hole in `unique`. This is the first place I looked when needing unique properties in irb, and only built the logic there once I realized `unique` wouldn't work for me.

I'm up for building the logic into `unique`, just seemed simpler to keep the logic separate since the function signatures are different. In that case we should probably follow the `options.transform` pattern used in `sorted`. lmk what you think

test/util/util.js
```diff
@@ -215,3 +217,106 @@ describe('insertAtIndex()', function() {
...
+    [1, 'abc', {}, null, undefined].forEach(input =>
+      expect(() => unique(input)).to.throwException()
+    );
+  });
+});
+
+describe('uniqueOfShallowObjects()', function() {
+  it('removes duplicate objects from an array', function() {
+    const keys = ['a', 'b'];
+    expect(JSON.stringify(uniqueOfShallowObjects([
```
evnp: You're totally right, I just assumed `eql` was strict equality (apparently `equal` is which is confusing...)

lib/util/array.js
tdumitrescu: I think my example rewrite of `unique` above shows that they're very much the same function, just with an optional transform (if that's what we want to call it; IDC too much one way or another).

evnp: 32fd5e0 I'd rather keep signatures consistent and I think there'd be too much overhead to changing `sorted` at this point, so let's do `transform`

lib/util/array.js
```diff
@@ -59,8 +59,21 @@ export function sorted(arr, options={}) {
...
+ * @param {function} [options.transform] - transforms each element for comparison
+ * Items are compared via strict equality so duplicate arrays/objects will not be removed
+ * unless a `transform` function is specified to hash objects before comparison.
+ * @example
+ * unique([{'a': 1, 'b': 2}, {'a': 1, 'b': 3}, {'a': 1, 'b': 2}], {transform: o => `${o['a']}:${o['b']}`});
+ * // [{'a': 1, 'b': 2}, {'a': 1, 'b': 3}]
+ */
+export function unique(arr, options={}) {
+  let transformed;
+  let encountered = new Set();
```
tdumitrescu: const

lib/util/array.js
```diff
@@ -59,8 +59,21 @@ export function sorted(arr, options={}) {
...
+  return arr.filter(item => {
+    transformed = options.transform ? options.transform(item) : item;
```
tdumitrescu: I don't care a whole lot about this, but `transformed` doesn't reflect this var's role very clearly to me and could also mislead someone into thinking they can do something like transform one object into another, which won't work; it's supposed to be a Key or Hash derived from the item.

test/util/util.js
```diff
@@ -215,3 +216,97 @@ describe('insertAtIndex()', function() {
...
+    ]);
+  });
+
+  it('throws an exception for non-array inputs', function() {
+    [1, 'abc', {}, null, undefined].forEach(input =>
+      expect(() => unique(input)).to.throwException()
+    );
+  });
+
+  it('removes duplicate objects from an array', function() {
```
tdumitrescu: Having one test case for "does not remove duplicate arrays or objects" and then another for "removes duplicate objects from an array" is very confusing. This case's docstring should really have something about using the transform function.

lib/util/array.js
evnp: I see what you mean. Let's just call it `hash`/`hashed`, you're right that it makes the intent more clear

PR #118, opened 2017-01-12
Add one line text mixin to common styles

jalack:


lib/stylesheets/mixins/common.styl
```diff
@@ -13,6 +13,11 @@ z-dropdown = 7000
 mp-box-shadow(size=10)
   box-shadow 0 (size)px (size * 4)px rgba(0, 0, 0, 0.15)
 
+one-line-text()
+  text-overflow ellipsis
+  white-space nowrap
+  overflow hidden
```
tdumitrescu: styles a-z (didn't @jbwyme add a stylus linter for this?)

tdumitrescu: might this be better in https://github.com/mixpanel/mixpanel-common/blob/36483bbcb087225aa01d5daa1762de4f098668e5/lib/stylesheets/mixins/typography.styl?

jbwyme: re: linter, there is one but turns out it's just throwing warnings for everything so I've got a branch in waiting that turns them all to errors

jbwyme: also i don't think we should use this mixin since we want to truncate middle everywhere

jalack: good to know!

PR #120, opened 2017-01-13
add an optional 'no-remove' attribute on mp-tag to make a tag without…

jhp:
… a remove action

### DIFF
--- a/lib/components/tag/index.jade
+++ b/lib/components/tag/index.jade
@@ -3,7 +3,7 @@
     .extra-icon
       svg-icon(attributes={icon: $component.getAttribute('icon')})
   .tag-name= $component.getAttribute('tag-name')
-  if !$component.getAttribute('no-remove')
+  if $component.isAttributeEnabled('removable')
     .remove-tag(onclick=$helpers.remove)
       .remove-icon
         svg-icon(attributes={icon: 'x'})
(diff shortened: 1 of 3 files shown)


lib/components/tag/index.jade
```diff
@@ -3,6 +3,7 @@
     .extra-icon
       svg-icon(attributes={icon: $component.getAttribute('icon')})
   .tag-name= $component.getAttribute('tag-name')
-  .remove-tag(onclick=$helpers.remove)
-    .remove-icon
-      svg-icon(attributes={icon: 'x'})
+  if !$component.getAttribute('no-remove')
```
jbwyme: what say we flip this logic and require removable tags to have `removable` set as an attribute?

tdumitrescu: prefer `$component.isAttributeEnabled('no-remove')` because it handles basic html boolean cases out of the box: https://github.com/mixpanel/webcomponent#built-in-helper-methods

tdumitrescu:
Sorry to merge-conflict you: since https://github.com/mixpanel/mixpanel-common/pull/110 has now landed in master, there are some differences in templating (using object notation for `class`, `attrs`, `props` etc). I can help you get this updated and converted.

PR #122, opened 2017-01-17
Add upsell popup

jalack:
We use these in various places and will need a new one for dashboards so adding a common component


lib/components/modal/index.styl
```diff
@@ -28,9 +28,6 @@
     opacity 0
     transform scale(1.1, 1.1)
 
-+host-styles(mp-modal)
-  width 400px
-
```
jalack: I'm a bit hesitant to remove this, but all the other popups & modals seemed to be working fine without it. @jbwyme let me know if there's an issue with removing it -- could accept width as an attribute or something

PR #125, opened 2017-01-18
boolean icon is not in fact just color one or color two...

tdumitrescu:
screen shot 2017-01-18 at 12 11 07 pm


jayjo:
The background color of the svg shape needs to be `grey-700` and the right dot needs to be `grey-900` - then that shit is stamped homie.

tdumitrescu:
The bg color isn't predefined for svg-icons, but right dot can be grey-900 fo sho

PR #126, opened 2017-01-18
Tutorial tooltip

evnp:
Used in https://github.com/mixpanel-platform/irb/pull/176
Mocks here https://app.zeplin.io/project.html#pid=5810e4117899d27c64470e4d&sid=586d4590ce6f17fe76c9d026
https://projects.invisionapp.com/d/main#/console/9497805/211293609/preview

### DIFF
--- a/examples/style-guide/index.jade
+++ b/examples/style-guide/index.jade
@@ -220,18 +220,19 @@
   .section.tooltips
     h1 Tooltips
 
+    h2 Standard tooltips
     .tooltips-container
       a.tooltip-container Default (placement 'top') tooltip
         mp-tooltip Tooltip above the element
 
       a.tooltip-container Placement 'bottom' tooltip
-        mp-tooltip(attrs={'placement': 'bottom'}) Tooltip below the element
+        mp-tooltip(attrs={placement: 'bottom'}) Tooltip below the element
 
       a.tooltip-container Placement 'left' tooltip
-        mp-tooltip(attrs={'placement': 'left'}) Tooltip on the left side of element
+        mp-tooltip(attrs={placement: 'left'}) Tooltip on the left side of element
 
       a.tooltip-container Placement 'right' tooltip
-        mp-tooltip(attrs={'placement': 'right'}) Tooltip on the right side of element
+        mp-tooltip(attrs={placement: 'right'}) Tooltip on the right side of element
 
       a.tooltip-container Multi-line tooltip
         mp-tooltip.multiline
@@ -240,6 +241,23 @@
       a.tooltip-container.position-relative Tooltip with non-statically positioned parent element
         mp-tooltip
           span A tooltip
+
+    h2 Tutorial tooltips
+    .tooltips-container
+      a.tutorial-tooltip-container.top.position-relative Default (placement 'top') tooltip
+        mp-tutorial-tooltip Tooltip above the element
+
+      a.tutorial-tooltip-container.bottom.position-relative Placement 'bottom' tooltip
+        mp-tutorial-tooltip(attrs={placement: 'bottom', numSteps: 4}) Tooltip below the element
+
+      a.tutorial-tooltip-container.left.position-relative Placement 'left' tooltip
+        mp-tutorial-tooltip(attrs={placement: 'left', numSteps: 4, currStep: 1}) Tooltip on the left side of element
+
+      a.tutorial-tooltip-container.right.position-relative Placement 'right' tooltip
+        mp-tutorial-tooltip(attrs={placement: 'right', numSteps: 4, currStep: 2}) Tooltip on the right side of element
+
+      a.tutorial-tooltip-container.center.position-relative Placement 'center' tooltip
+        mp-tutorial-tooltip(attrs={placement: 'center', overrideTop: '20px', numSteps: 4, currStep: 3}) Tooltip centered on the element
 
   .section.widgets
     h1 Widgets
--- a/examples/style-guide/index.jade
+++ b/examples/style-guide/index.jade
@@ -220,7 +220,6 @@
   .section.tooltips
     h1 Tooltips
 
-    h2 Standard tooltips
     .tooltips-container
       a.tooltip-container Default (placement 'top') tooltip
         mp-tooltip Tooltip above the element
@@ -241,23 +240,6 @@
       a.tooltip-container.position-relative Tooltip with non-statically positioned parent element
         mp-tooltip
           span A tooltip
-
-    h2 Tutorial tooltips
-    .tooltips-container
-      a.tutorial-tooltip-container.top.position-relative Default (placement 'top') tooltip
-        mp-tutorial-tooltip Tooltip above the element
-
-      a.tutorial-tooltip-container.bottom.position-relative Placement 'bottom' tooltip
-        mp-tutorial-tooltip(attrs={placement: 'bottom', numSteps: 4}) Tooltip below the element
-
-      a.tutorial-tooltip-container.left.position-relative Placement 'left' tooltip
-        mp-tutorial-tooltip(attrs={placement: 'left', numSteps: 4, currStep: 1}) Tooltip on the left side of element
-
-      a.tutorial-tooltip-container.right.position-relative Placement 'right' tooltip
-        mp-tutorial-tooltip(attrs={placement: 'right', numSteps: 4, currStep: 2}) Tooltip on the right side of element
-
-      a.tutorial-tooltip-container.center.position-relative Placement 'center' tooltip
-        mp-tutorial-tooltip(attrs={placement: 'center', overrideTop: '20px', numSteps: 4, currStep: 3}) Tooltip centered on the element
 
   .section.widgets
     h1 Widgets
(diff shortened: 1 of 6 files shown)


lib/widgets/tutorial-tooltip/index.styl
```diff
@@ -0,0 +1,110 @@
...
+  border-radius 8px
+  box-shadow 0 2px 4px 0 (rgba(0, 0, 0, 0.16))
+  color white
+  cursor default
+  font-family family-helvetica
+  font-size-default()
+  height tooltip-height
+  position absolute
+  width tooltip-width
+  z-index 1
```
jordanmnunez: shouldn't this *always* be on top. is z-index 1 enough?

evnp: This was enough for irb but you're right. DropMenu component uses z-index 100, I'll go with that

jordanmnunez: classic 99999999999

evnp: Over 9000!

lib/widgets/tutorial-tooltip/index.jade
```diff
@@ -0,0 +1,10 @@
+.mp-tutorial-tooltip-wrapper(
+  class={[`placement-${placement}`]: placement}
+  style=overrideStyles
```
tdumitrescu: a few points:
- this override seems to mirror the functionality of the existing [`style-override` attr](https://github.com/mixpanel/panel/blob/1552b13b2b3b5a6c8882e577bd61df296ce9bba0/lib/component.js#L257) pretty closely
- @jbwyme and I are in agreement that this is a pattern we want to avoid if at all possible; widget/components should expose pre-defined style options (like what you have with the `placement` attr here), with arbitrary styling only as a last resort for exceptional situations

examples/style-guide/index.jade
```diff
@@ -241,6 +242,23 @@
         mp-tooltip
           span A tooltip
 
+    h2 Tutorial tooltips
```
tdumitrescu: If this is a 'widget' rather than a base component, then it should go into the bottom section of the style guide

lib/widgets/tutorial-tooltip/index.js
```diff
@@ -0,0 +1,56 @@
...
+    };
+  }
+
+  attachedCallback() {
+    super.attachedCallback();
+
+    let update = {};
+
+    if (this.hasAttribute(`placement`)) {
+      update.placement = this.getAttribute(`placement`);
```
tdumitrescu: It's generally a good pattern to put attribute-reading code into `attributeChangedCallback`, which also handles post-attachment updates. This element is kind of badly behaved in that regard.

lib/widgets/tutorial-tooltip/index.js
```diff
@@ -0,0 +1,56 @@
...
+    }
+
+    const overrides = {
+      top: this.getAttribute(`overrideTop`),
```
tdumitrescu: nit: attrs should be dashed: `override-top` etc

lib/widgets/tutorial-tooltip/index.js
```diff
@@ -0,0 +1,56 @@
...
+    };
+
+    [`top`, `bottom`, `left`, `right`].forEach(attr => {
+      if (overrides[attr]) {
+        update.overrideStyles = update.overrideStyles || {};
+        update.overrideStyles[attr] = overrides[attr];
+      }
+    });
+
+    [`numSteps`, `currStep`].forEach(attr => {
```
tdumitrescu: `num-steps`, `cur-step`

lib/widgets/tutorial-tooltip/index.styl
```diff
@@ -0,0 +1,110 @@
+@import '../../stylesheets/defaults'
+
+tooltip-width = 420px
+tooltip-height = 160px
+tooltip-distance = 12px
```
tdumitrescu: I think the 'new thing' is to start prefixing stylus vars with $. we'll have to backport existing stylesheets but might as well be doing it for new ones.

lib/widgets/tutorial-tooltip/index.styl
```diff
@@ -0,0 +1,110 @@
...
+
+.mp-tutorial-tooltip-wrapper
```
tdumitrescu: Is there going to be any intro/exit animation for these? Not blocking this PR for sure, but when it comes time I can help get you up to speed with the current implementation possibilities.

examples/style-guide/index.jade
evnp: gotcha, didn't catch that

lib/widgets/tutorial-tooltip/index.styl
evnp: I'm :100:% on board with this

lib/widgets/tutorial-tooltip/index.styl
evnp: There probably will, going to work with Jeff on that once this is live in his hands. I'll bug you about this then, thanks

lib/widgets/tutorial-tooltip/index.jade
evnp: I did this for the sake of flexibility, but I can see why it's not preferred. For context, the reason I need this is because in one of the panes the tooltip is centered horizontally over its container, but offset by a px value from the top - doesn't make sense to code that px value into this widget.

The external interface is actually through `overrideTop`, `overrideLeft`, etc. attrs, is that closer to what you meant by a "pre-defined style option"? They get turned into this `overrideStyles` object internally for simplicity of applying in the template, but that doesn't have to happen. Happy to take any suggestions on how this should work, I'm still getting aquatinted with best practices here

tdumitrescu: Yeah I misread this `overrideStyles` thing as being an arbitrary style override, rather than pre-composed out of a limited set. That's definitely better, though I still hate making the user of this component put css into markup. @jbwyme has been doing some thinking around tooltips and will be integrating the tutorial-tooltip into dashboard as well, so he'll chime in with a more "pure css" solution

jbwyme: Yes we can do better! You can use the `host-styles` mixin to specify the positions in css on the host element. Using a selector for the attributes allows you to have different positions. Then that allows the client to override directly using css.

Here's a similar example: 

https://github.com/mixpanel/mixpanel-common/blob/202162a090238a2eaf1f72679c2f7117589efd12/lib/components/list-item/index.styl#L11

evnp: This sounds fantastic, I'm still a little unclear from the example how it works though. Is there an example of a client overriding using `host-styles` I can look at? 
```

+host-styles(mp-list-item, ':first-child')
  margin-top $mp-list-gutter-height
```
In my case `':first-child'` would be the placement class? How are the styles from the client being passed in?

lib/widgets/tutorial-tooltip/index.styl
```diff
@@ -1,41 +1,42 @@
...
++host-styles(mp-tutorial-tooltip, '[placement="right"]')
+  left "calc(100% + %s)" % tooltip-distance
+  top "calc(50% - %s)" % (tooltip-height / 2)
 
-  &.placement-right
-    left "calc(100% + %s)" % tooltip-distance
-    top "calc(50% - %s)" % (tooltip-height / 2)
++host-styles(mp-tutorial-tooltip, '[placement="center"]')
+  left "calc(50% - %s)" % (tooltip-width / 2)
+  top "calc(50% - %s)" % (tooltip-height / 2)
```
jbwyme: woo there we go

lib/widgets/tutorial-tooltip/index.styl
```diff
@@ -1,41 +1,42 @@
...
-  box-shadow 0 2px 4px 0 (rgba(0, 0, 0, 0.16))
-  color white
-  cursor default
-  font-family family-helvetica
-  font-size-default()
++host-styles(mp-tutorial-tooltip)
   height tooltip-height
   position absolute
   width tooltip-width
   z-index 100
```
jbwyme: yayayaya

examples/style-guide/index.styl
```diff
@@ -136,3 +136,30 @@
...
+          margin 100px
+
+        &.right
+          margin 100px
+
+        &.center
+          margin-bottom 200px
+
+          mp-tutorial-tooltip
+            top 40px
```
tdumitrescu: yussssss

lib/widgets/tutorial-tooltip/index.js
```diff
@@ -0,0 +1,37 @@
...
+    super.attributeChangedCallback(...arguments);
+
+    let update = {};
+    const placement = this.getAttribute(`placement`);
+    const numSteps = parseInt(this.getAttribute(`num-steps`));
+    const currStep = parseInt(this.getAttribute(`curr-step`));
+
+    if (placement) { update.placement = placement; }
+    if (Number.isInteger(numSteps)) { update.numSteps = numSteps; }
+    if (Number.isInteger(currStep)) { update.currStep = currStep; }
```
tdumitrescu: I gotta admit, I'm kinda shocked we don't have [One True Brace Style](http://eslint.org/docs/rules/brace-style#tbs) enforced by the linter. I guess it's rarely broken around here! But if you'd like to save the trouble before it gets added to the linter config, you could multiline these.............

PR #127, opened 2017-01-19
Sentence-casing tags

jalack:


lib/util/string.js
```diff
@@ -53,3 +53,8 @@ export function stringFilterMatches(str, filterStr) {
 
   return matches;
 }
+
+export function toSentenceCase(string) {
```
jbwyme: can we add a test for this?

lib/components/tag/index.jade
```diff
@@ -5,7 +5,7 @@
   if $component.getAttribute('icon')
     .extra-icon
       svg-icon(attrs={icon: $component.getAttribute('icon')})
-  .tag-name= $component.getAttribute('tag-name')
+  .tag-name= $helpers.toSentenceCase($component.getAttribute('tag-name'))
```
jbwyme: my first thought is that this is too business-logicy to put in the component but then I figure let's let the first use-case where they DON'T want to sentence case deal with it.

examples/style-guide/index.js
```diff
@@ -34,8 +34,8 @@ document.registerElement('style-guide', class extends Component {
         selectedBookmarkId: null,
         tagSelectorLoadState: 'idle',
         tagSelectorData: {
-          selectedTags: new Set(["My tag", "Another tag"]),
-          allTags: new Set(["My tag", "Another tag", "Our tag", "His tag"]),
```
jbwyme: deliberately not lowercasing `Our tag` and `His tag` ?

lib/components/tag/index.js
```diff
@@ -17,6 +18,7 @@ registerMPElement(`mp-tag`, class extends Component {
           e.stopPropagation();
           this.dispatchEvent(new CustomEvent(`change`, {detail: {action: `remove`}}));
         },
+        toSentenceCase: tagName => toSentenceCase(tagName),
```
jbwyme: `toSentenceCase,` should suffice

lib/widgets/tag-selector/index.jade
```diff
@@ -51,9 +51,9 @@
           )
             for token, pos in $helpers.stringFilterMatches(tag, inputText)
               if pos % 2 === 1
-                span= token
+                span= $helpers.toSentenceCase(token)
```
jbwyme: this will sentence case substrings, I think you want to call this on line 52 around `tag`

lib/widgets/tag-selector/index.js
```diff
@@ -40,6 +40,7 @@ registerMPElement(`mp-tag-selector`, class extends Component {
             && !this.state.selectedTags.has(this.state.inputText);
           return isNewTag && !this.isAttributeEnabled(`read-only`);
         },
+        toSentenceCase: tagName => toSentenceCase(tagName),
```
jbwyme: `toSentenceCase,`

lib/components/tag/index.jade
jalack: I went back & forth on this a lot, whether the consumer should just deal with this and format accordingly or whether it should be considered a part of how the component renders. like you said though, in every case I know of we want this behavior, so I'll lean towards just having it work as we want out of the box & consider decoupling as needed later

lib/components/tag/index.js
jalack: oh very cool, didn't know about that trick

examples/style-guide/index.js
jalack: not really, I mostly was just trying to validate on different inputs. given that our tags will be lowercase though, will change them around

test/util/string.js
```diff
@@ -1,55 +1,84 @@
...
+    expect(toSentenceCase(`a`)).to.eql(`A`);
+  });
+
+  it(`converts all letters but the first to lowercase`, function() {
+    expect(toSentenceCase(`CHOCOLATE CAKE`)).to.eql(`Chocolate cake`);
+  });
+
+  it(`produces the empty string on null input`, function() {
+    expect(toSentenceCase(null)).to.eql(``);
+  });
```
jbwyme: sweet tests

test/widgets/tag-selector.js
```diff
@@ -22,36 +22,36 @@ describe('Test tag-selector widget', () => {
...
-  it('fires change event to remove tags', function(done) {
-    this.widget.addEventListener('change', e => {
-      expect(e.detail.action).to.equal('removeTag');
+  it(`fires change event to remove tags`, function(done) {
+    this.widget.addEventListener(`change`, e => {
+      expect(e.detail.action).to.equal(`removeTag`);
       done();
     });
-    this.widget.el.querySelector('mp-tag').dispatchEvent(new CustomEvent('change', {detail: {action: 'remove'}}));
+    this.widget.el.querySelector(`mp-tag`).dispatchEvent(new CustomEvent(`change`, {detail: {action: `remove`}}));
```
jbwyme: i don't think our linter currently runs against the tests but probably should

jalack: my editor bitches at me when stuff isn't linted and auto-fixes easy problems so I've just been tidying up as I go

PR #129, opened 2017-01-19
upgrade panel for new ID behavior

tdumitrescu:
Running panel code from two separate copies (as we sometimes do
with mixpanel-common components + app code) could cause ID collisions
because it was a simple auto-incrementing counter scoped to the
module. This version uses collision-resistent IDs and also turns
them into internal component props rather than HTML attrs.


tdumitrescu:
Thx, published in v0.37.0

PR #132, opened 2017-01-21
Add pinned tags that stay on top of list & default tags that cannot be removed

jalack:
The use case for this is the “All Reports” tag in dashboards. Pinned tags stay at the top of the list, but then go away once the user starts typing. Default tags would include our `segmentation` tag -- it can be seen but not removed by the user. My two main concerns here are that this might be a bit specific for a common component, and that the uber-functional style of this code might be a bit esoteric. Would love to hear thoughts from folks.

### DIFF
--- a/lib/util/function.js
+++ b/lib/util/function.js
@@ -14,13 +14,17 @@ export function mapArguments(fn, argMap) {
 }
 
 /**
- * compose two orderings lexically to create a single combined ordering.
+ * Compose an arbitrary number of orderings lexically to create a single combined ordering.
  */
-export function lexicalCompose(fn1, fn2) {
+export function lexicalCompose(...args) {
   return (a, b) => {
-    let cmp = fn1(a, b);
-    return cmp === 0 ? fn2(a, b) : cmp;
+    const order = idx => {
+      const cmp = args[idx](a, b);      
+      return cmp === 0 && idx !== args.length - 1 ? order(++idx) : cmp;
+    };
+    return order(0);
   };
 }
+
 
 export const defaultOrdering = leqToNumericOrdering((a, b) => a <= b);
--- a/lib/util/function.js
+++ b/lib/util/function.js
@@ -17,7 +17,7 @@ export function mapArguments(fn, argMap) {
  * Compose an arbitrary number of orderings lexically to create a single combined ordering.
  */
 export function lexicalCompose(...args) {
-  return (a, b) => args.reduce((acc, curr) => acc === 0 ? curr(a, b) : acc);
+  return (a, b) => args.reduce((acc, curr) => acc === 0 ? curr(a, b) : acc, 0);
 }
 
 
--- a/lib/util/function.js
+++ b/lib/util/function.js
@@ -17,13 +17,7 @@ export function mapArguments(fn, argMap) {
  * Compose an arbitrary number of orderings lexically to create a single combined ordering.
  */
 export function lexicalCompose(...args) {
-  return (a, b) => {
-    const order = idx => {
-      const cmp = args[idx](a, b);      
-      return cmp === 0 && idx !== args.length - 1 ? order(++idx) : cmp;
-    };
-    return order(0);
-  };
+  return (a, b) => args.reduce((acc, curr) => acc === 0 ? curr(a, b) : acc);
 }
 
 
--- a/lib/widgets/tag-selector/index.js
+++ b/lib/widgets/tag-selector/index.js
@@ -81,6 +81,7 @@ registerMPElement(`mp-tag-selector`, class extends Component {
     super.attachedCallback(...arguments);
     this.update({
       pinnedTags: new Set(this.getJSONAttribute(`pinned-tags`)),
+      defaultTags: new Set(this.getJSONAttribute(`default-tags`)),
       selectedTags: new Set(this.getJSONAttribute(`selected-tags`)),
       allTags: new Set(this.getJSONAttribute(`all-tags`)),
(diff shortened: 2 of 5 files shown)


lib/util/function.js
```diff
@@ -14,13 +14,17 @@ export function mapArguments(fn, argMap) {
...
   return (a, b) => {
-    let cmp = fn1(a, b);
-    return cmp === 0 ? fn2(a, b) : cmp;
+    const order = idx => {
+      const cmp = args[idx](a, b);      
+      return cmp === 0 && idx !== args.length - 1 ? order(++idx) : cmp;
+    };
+    return order(0);
   };
 }
```
jhp: The one problem with this pattern is that it grows the stack, since JS doesn't have tail-call elimination. It wouldn't be a problem in practice here because the numbers involved are small, but you might want to get in the habit of avoiding it.

I'd recommend either using a plain for/while loop, or a higher-order function like `reduce`.

jhp:
I'm going to write tests for util/function.js today, so I'll cover the updated API of lexicalCompose as well.

jalack:
I agree that this does feel like starting down a potentially dangerous path in terms of bloating each component with too much application-specific functionality. Maybe this feature is useful elsewhere, maybe it isn't, but I'm definitely open to suggestions on how we can get what we want from the `mixpanel-common` component in a cleaner way. I don't like the idea of forking everything either, but it's definitely a balance we need to strike

lib/util/function.js
jalack: awesome tip, thanks!

lib/util/function.js
```diff
@@ -14,13 +14,11 @@ export function mapArguments(fn, argMap) {
...
- * compose two orderings lexically to create a single combined ordering.
+ * Compose an arbitrary number of orderings lexically to create a single combined ordering.
  */
-export function lexicalCompose(fn1, fn2) {
-  return (a, b) => {
-    let cmp = fn1(a, b);
-    return cmp === 0 ? fn2(a, b) : cmp;
-  };
+export function lexicalCompose(...args) {
+  return (a, b) => args.reduce((acc, curr) => acc === 0 ? curr(a, b) : acc);
```
jbwyme: baller

lib/widgets/tag-selector/index.js
```diff
@@ -118,17 +120,22 @@ registerMPElement(`mp-tag-selector`, class extends Component {
...
+      .filter(tag => !search.length || !this.state.pinnedTags.has(tag))
       .filter(tag => !this.state.selectedTags.has(tag) && stringFilterMatches(tag, search))
       .sort(
         lexicalCompose(
-          // first sort by whether the string matches the search string exactly
-          mapArguments(defaultOrdering, a => a.toLowerCase() === search ? 0 : 1),
-          // then sort alphabetically
-          mapArguments(defaultOrdering, a => a.toLowerCase())));
+          // put the pinned tags on top
+          mapArguments(defaultOrdering, a => this.state.pinnedTags.has(a) ? 0 : 1),
```
jbwyme: let's swap pinned and search so that an exact search match is still at the top

lib/widgets/tag-selector/index.js
```diff
@@ -80,6 +80,8 @@ registerMPElement(`mp-tag-selector`, class extends Component {
   attachedCallback() {
     super.attachedCallback(...arguments);
     this.update({
+      pinnedTags: new Set(this.getJSONAttribute(`pinned-tags`)),
+      defaultTags: new Set(this.getJSONAttribute(`default-tags`)),
```
jbwyme: can you add these to the `defaultState` as well and add a comment about what their purpose is (defaultTags cannot be removed for instance).

jbwyme:
Re: the conversation about mixpanel-common and the level of abstraction that we should focus on: I think `components` should be generic; `widgets` can be much more business-logicy. If someone needs to use a modified version of a widget down the road they can either refactor and make it a little more generic to fit their use-case OR they can break it into some smaller components that are useful to building their own widget. In this particular case, we just need to share it between two code-bases so having it as a widget seemed like a reasonable trade-off. I'm totally fine with making it tightly coupled to the needs of our app right now.

lib/widgets/tag-selector/index.js
```diff
@@ -118,17 +124,22 @@ registerMPElement(`mp-tag-selector`, class extends Component {
     const search = this.state.inputText.trim().toLowerCase();
     const matchingTags = [...this.state.allTags]
       .map(tag => tag.toLowerCase())
+      .filter(tag => !search.length || !this.state.pinnedTags.has(tag))
```
jbwyme: hmm do you think pinnedTags should not be shown in the autocomplete results?

jalack: the use case is `All reports` which we don't want to show once the user starts typing

jbwyme:
ahh, i think we may end up making some changes to how All reports works but
this is fine for now until we know what those are.

lib/util/function.js
jhp: I think you need a seed argument of 0 here (second argument to reduce).

jalack: if not provided, should be `undefined` right? and `undefined !== 0`

jhp: I think if not provided then the first element of the array is used as a seed, and reduction starts with the second argument.

jbwyme:
just a follow-up here: let's leave this PR hanging out for now, we may not end up needing it at all #sucker

PR #134, opened 2017-01-23
cleaned up styleguide example for bookmarks

rickydoar:
cleaned up the bookmark opens and added headers for new bookmark examples


tdumitrescu:
published v0.38

PR #135, opened 2017-01-23
box shadow re: Mason

rickydoar:
added the correct drop shadow

### DIFF
--- a/lib/widgets/upsell-modal/index.styl
+++ b/lib/widgets/upsell-modal/index.styl
@@ -7,6 +7,7 @@ mp-modal
   .content
     display flex
     padding 20px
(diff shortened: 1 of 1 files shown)


lib/widgets/upsell-modal/index.styl
```diff
@@ -7,6 +7,7 @@ mp-modal
   .content
     display flex
     padding 20px
+    box-shadow 0 1px 30px 0 rgba(0, 0, 0, 0.19)
```
jalack: just to be difficult: could use `alpha(black, 0.19)` here 😄

tdumitrescu:
v0.38.1

PR #136, opened 2017-01-25
mp-toggle updates

tdumitrescu:
- choose styles via attrs instead of proxying CSS classes
- added super square mode

screen shot 2017-01-24 at 4 20 38 pm


lib/components/toggle/index.js
```diff
@@ -29,6 +27,9 @@ registerMPElement(`mp-toggle`, class extends Component {
   attributeChangedCallback(attr, oldVal, newVal) {
     super.attributeChangedCallback(...arguments);
     if (attr === `selected`) {
+      this.dispatchEvent(new CustomEvent(`select`, {detail: {selected: newVal}}));
```
jbwyme: yesssss

PR #140, opened 2017-01-26
Add mp toast

jalack:
Please be gentle


lib/components/toast/index.js
```diff
@@ -0,0 +1,109 @@
...
+registerMPElement(`mp-toast`, class extends Component {
+
+  get config() {
+    return {
+      css,
+      template,
+      useShadowDom: true,
+      defaultState: {
+        toasts: {},
+        openTime: 3000,
```
tdumitrescu: this is supposed to control how long it stays open? We definitely need an option for "no auto close" which is what IRB has.

lib/components/toast/index.js
```diff
@@ -0,0 +1,109 @@
...
+      },
+    };
+  }
+
+  attributeChangedCallback(attr) {
+    super.attributeChangedCallback(...arguments);
+    if (this.initialized && attr === `toasts`) {
+      const toasts = this.getJSONAttribute(`toasts`);
+      
+      this._toastsToClose = this._toastsToClose.filter(toast => toast.id in toasts);
```
tdumitrescu: I feel like this need for our own little diffing algorithm that understands which toasts need to animate out is an unfortunate effect of the "list of toasts in a single toast" API. Seems pretty clever and pretty fragile.

lib/components/toast/index.styl
```diff
@@ -0,0 +1,79 @@
+@import '../../stylesheets/defaults'
+
+@keyframes fadeIn
```
tdumitrescu: why do we need keyframe animations for a simple margin + opacity transition?

PR #142, opened 2017-01-26
chart analysis type utils

rickydoar:
added logarithmic and rolling data transformation utils


tdumitrescu:
IRB

rickydoar:
Changed to match IRB rolling average. Logarithmic is done by Highcharts.

tdumitrescu:
Is this exactly the same as in IRB? what can/should we extract to common as a result?

rickydoar:
The logic is identical. I changed variable names slightly so it was more readable as to what was happening.

tdumitrescu:
So now we have two almost-but-not-quite-the-same versions?

rickydoar:
I would assume IRB would start using the mp-common utils here, change nothing else about their code and then we'd have only one version.

tdumitrescu:
That's fine - can you open an IRB PR to swap it out?

rickydoar:
The functions are identical with `window` => `values` and `windowSize` => `rollingCount`, since the previous variable names didn't seem very intuitive. I can change it back if you feel strongly about it?

rickydoar:
Yup can do. Can I do it after this one is merged?

lib/util/index.js
```diff
@@ -180,3 +180,57 @@ export function truncateMiddle(string, len) {
...
+    case `day`:
+      return 7;
+    case `week`:
+      return 5;
+    case `month`:
+      return 3;
+  }
+}
+
+export function mapObjectKeys(obj, callback) {
```
tdumitrescu: is this different from https://github.com/mixpanel/mixpanel-common/blob/master/lib/util/index.js#L103-L106?

tdumitrescu:
Sure but how about before that you also copy the IRB tests for `nestedObjectRolling` so we can have some confidence that they behave the same?

lib/util/index.js
tdumitrescu: or uhh was it actually this one https://github.com/mixpanel/mixpanel-common/blob/master/lib/util/index.js#L108-L111

rickydoar: It is different, but it's not different than https://github.com/mixpanel/mixpanel-common/blob/master/lib/util/index.js#L108-L111 😄

tdumitrescu: 4 minutes too late homie

rickydoar: hah got me

PR #143, opened 2017-01-27
first pass at toaster

jbwyme:
http://cometapp.us/c/XRZak03y


lib/components/toast/index.js
```diff
@@ -0,0 +1,21 @@
...
+import css from './index.styl';
+
+registerMPElement(`mp-toast`, class extends Component {
+  get config() {
+    return {
+      css,
+      template,
+      useShadowDom: true,
+      helpers: {
+        ctaClick: () => this.dispatchEvent(new CustomEvent(`action`)),
```
tdumitrescu: Call me old-fashioned, but don't we want cta click to fire a close event as well? Otherwise the app code has to remember to handle it explicitly as part of the `action` handler.

examples/style-guide/index.jade
```diff
@@ -229,6 +229,22 @@
...
+    button(on={click: $helpers.addToast}) Add some toast
+    .toast-container
+      for toast in toasts
+        mp-toast(
+          attrs={
+            cta: toast.cta,
+            hide: toast.closed,
+          }
+          on={
+            action: () => alert('undo please'),
```
tdumitrescu: byebye forever `onchange`

examples/style-guide/index.js
```diff
@@ -126,6 +127,14 @@ document.registerElement('style-guide', class extends Component {
           this.state.open.tagSelector = false;
           this.update();
         },
+        addToast: () => {
+          this.state.toasts.push({message: 'Hi, I\'m a toast', cta: 'undo', closed: false});
```
tdumitrescu: - you know what would keep you from needing to slash-escape that quote? `strictly better`
- more pertinently, I think it'd make sense to give each toast a different message, to ensure that the UI is really doing what you want (i.e. the vdom differ isn't just updating attrs on an existing toast when one is deleted/inserted). The animations show that it's OK but it'd be a lot clearer with different content too.

lib/components/toast/index.jade
```diff
@@ -0,0 +1,12 @@
+.mp-toast-stage
+  if !$component.isAttributeEnabled('hide')
+    .mp-toast-container(class={
+        delayed: {'mp-toast-visible': true},
+        remove: {'mp-toast-out': true, delayRemove: 250},
+      })
```
tdumitrescu: our standard jade indentation would be
```jade
.mp-toast-container(class={
  delayed: {'mp-toast-visible': true},
  remove: {'mp-toast-out': true, delayRemove: 250},
})
```

examples/style-guide/index.styl
```diff
@@ -80,6 +80,12 @@
         mp-tooltip.multiline span
           max-width 200px
 
+  .toast-container
```
tdumitrescu: So is it the case that apps are expected to set aside a "landing area" for these toasts? ie no automatic absolute positioning etc

examples/style-guide/index.js
jbwyme: yeah sgtm

examples/style-guide/index.styl
jbwyme: Yeah my intuition is that you'll need a container of some sort to position a group of toasts and i didn't want this component to be that heavy. As we use these and a generic pattern develops I'm happy to either a) revisit or b) build a "toast widget" or what-have-you.

lib/components/toast/index.js
jbwyme: I don't feel strongly either way. My reason for leaving it out was to avoid making assumptions about what should happen when the action was clicked but I can't really think of a reasonable case where we wouldn't want to close the toast.

jbwyme: i'll make it respect `e.preventDefault`

jbwyme: https://github.com/mixpanel/mixpanel-common/pull/143/files#diff-983bb00753f5e06c3eb17103da043f77R26

PR #149, opened 2017-01-30
Tutorial tooltip styles

evnp:

PR #150, opened 2017-01-31
Tutorial tooltip updates

jhp:
- Add "hide-footer" attribute to hide the footer
- Add "arrow-offset" attribute to override the position of the small arrow (which is centered by default)
- Add new event to tag-selector on input change. This is to support showing a tooltip next to the "my dashboard" tag.
- Add 'tag-name' attribute to mp-list-items in tag-selector. This is so that the list item for a tag can be found conveniently with a DOM query.
- Add "hide" attribute to hide the tutorial-tooltip

### DIFF
--- a/lib/widgets/tutorial-tooltip/index.jade
+++ b/lib/widgets/tutorial-tooltip/index.jade
@@ -6,4 +6,4 @@
       ul.steps
         each step in $helpers.getSteps()
           li.step(class={active: step === currStep})
-  .arrow(class={ [`arrow-align-${arrowAlign}`]: arrowAlign})
+  .arrow(class={[`arrow-align-${arrowAlign}`]: arrowAlign})
--- a/lib/widgets/tutorial-tooltip/index.js
+++ b/lib/widgets/tutorial-tooltip/index.js
@@ -43,7 +43,7 @@ registerMPElement(`mp-tutorial-tooltip`, class extends Component {
     if (Number.isInteger(currStep)) {
       update.currStep = currStep;
     }
-    if(arrowAlign) {
+    if (arrowAlign) {
       update.arrowAlign = arrowAlign;
     }
 
--- a/lib/widgets/tutorial-tooltip/index.styl
+++ b/lib/widgets/tutorial-tooltip/index.styl
@@ -43,7 +43,6 @@ $tooltip-distance = 12px
   cursor default
   height inherit
 
-    
   &:not(.placement-center)
     .arrow
       border 5px solid transparent


tdumitrescu:
@evnp @jhp do you two need to coordinate here? https://github.com/mixpanel/mixpanel-common/pull/149

jhp:
@evnp yes we can follow your design, I just needed to be able to hide the tooltip so I added an attribute, but it doesn't support animations so we should go with your setup.

lib/widgets/tutorial-tooltip/index.js
```diff
@@ -38,6 +45,29 @@ registerMPElement(`mp-tutorial-tooltip`, class extends Component {
...
+      switch(this.state.placement) {
+        case `top`:
+        case `bottom`:
+          arr.style.left = (arrowOffset - 5) + `px`;
+          break;
+        case `left`:
+        case `right`:
+          arr.style.top = (arrowOffset - 5) + `px`;
+      }
+    }
```
evnp: I'm not a huge fan of doing this via js. Would be great to keep styles in stylus - seems like a use case for [CSS custom properties](https://developer.mozilla.org/en-US/docs/Web/CSS/Using_CSS_variables) described here https://developers.google.com/web/fundamentals/getting-started/primers/shadowdom

screen shot 2017-01-31 at 9 28 53 am

jhp: it looks like these are not supported in IE. https://developer.mozilla.org/en-US/docs/Web/CSS/Using_CSS_variables

tdumitrescu: As usual, let me request that we try as much as possible NOT to open up avenues for completely arbitrary styling, but stick to a limited set of pre-made style options. (i.e. prefer `<my-widget arrow-left>` to `<my-widget arrow-offset-x="15">`, and likewise with the css equivalents)

tdumitrescu: yeah css vars are still too patchily supported for us to rely on

jhp: @tdumitrescu why do you make that request? I would say that making a component more abstract by having more-arbitrary parameters is generally better.

tdumitrescu: Because
- it encourages design deviations. "like this other existing widget, but 3px taller and a smidge darker" has been the scourge of our design language in the past. The design team is now on board with enforcing consistency across the product, and allowing arbitrary styling works against that.
- not allowing arbitrary CSS limits the leaking of abstractions into the clients of these components. so when you change the classname of an internal div within the component's markup, you don't accidentally bork styling in some random client app which was depending on overriding styles associated with that class. Using an `attr` for this does generally prevent that kind of issue since it's explicitly part of the public API, but then you're using attrs for styling i.e. basically inline CSS blech.

That said we do expose some 'host overrides' like the positioning attrs on these tooltips, but it should be the exception rather than the rule.

evnp: Sad, no polyfill for that? I'm with ted, for style consistency and keeping styles in stylesheets where they belong I'd prefer a set of arrow style preset attrs that translate to a class on the element that aligns the arrow in stylus. What about an `arrow-align` attr with string values `"top", "right", "bottom", "left"` exactly like the `placement` attr?

This is assuming you need tooltips like this (arrow align top, center, bottom) and not arbitrary arrow positions in between.
```
   /    /    /
 <     |    |
  |    |   <
  |   <     |
   \    \    \
```

lib/widgets/tutorial-tooltip/index.styl
```diff
@@ -51,6 +52,9 @@ $tooltip-distance = 12px
       position absolute
       width 0
 
+      &.arrow-align-top.arrow-align-top
+        top 23px
+
```
evnp: If you want to push this out as-is go for it. I'd like to expand this to handle all the various arrow position cases, can do it in https://github.com/mixpanel/mixpanel-common/pull/149/. I also think this makes more sense on the same level as `.placement-xyz` so that we can go back to the psuedoelement and  avoid the double `.arrow-align-top.arrow-align-top` above

lib/widgets/tutorial-tooltip/index.js
```diff
@@ -44,30 +44,12 @@ registerMPElement(`mp-tutorial-tooltip`, class extends Component {
     if (Number.isInteger(currStep)) {
       update.currStep = currStep;
     }
+    if(arrowAlign) {
```
evnp: sorry to nitpick, space after `if`

lib/widgets/tutorial-tooltip/index.jade
```diff
@@ -6,4 +6,4 @@
       ul.steps
         each step in $helpers.getSteps()
           li.step(class={active: step === currStep})
-  .arrow
+  .arrow(class={ [`arrow-align-${arrowAlign}`]: arrowAlign})
```
evnp: more nit, no space between `{ [`

examples/style-guide/index.jade
```diff
@@ -362,7 +362,7 @@
     h2 Tutorial tooltip widgets
     .tutorial-tooltips-container
       a.tutorial-tooltip-container.top.position-relative Default (placement 'top') tooltip
-        mp-tutorial-tooltip(attrs={'hide-footer': true, 'arrow-offset': 20}) Tooltip above the element
+        mp-tutorial-tooltip(attrs={'hide-footer': true}) Tooltip above the element
```
evnp: Would be good to add cases for arrow-align here - this styleguide acts as documentation for the widgets. If you don't have time I'll add in https://github.com/mixpanel/mixpanel-common/pull/149

lib/widgets/tutorial-tooltip/index.js
```diff
@@ -20,13 +20,20 @@ registerMPElement(`mp-tutorial-tooltip`, class extends Component {
     };
   }
 
+  attachedCallback() {
+    super.attachedCallback(...arguments);
+    this.style.display = this.isAttributeEnabled(`hide`) ? `none` : ``;
```
evnp: https://github.com/mixpanel/mixpanel-common/pull/149 handles this for you yeah? if you need this right now go ahead and push, and I'll remove in that branch if that works

evnp: and by that I mean lmk when you're ready and I'll merge+publish

jhp: okay can you please merge+publish?

evnp: will do as soon as build passes

PR #163, opened 2017-02-09
Play nice with Firefox & Safari

jalack:
Fixing a couple issues that arose with cross-browser testing

### DIFF
--- a/lib/components/tag/index.styl
+++ b/lib/components/tag/index.styl
@@ -3,6 +3,7 @@
 .mp-tag
   background-color grey-700
   border-radius 13px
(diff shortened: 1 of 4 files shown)


lib/components/tag/index.styl
```diff
@@ -3,6 +3,7 @@
 .mp-tag
   background-color grey-700
   border-radius 13px
+  box-sizing content-box
```
jbwyme: can we just use border-box and fix the styles instead?

tdumitrescu:
published 0.46.2

PR #164, opened 2017-02-11
updated the square toggle

jayjo:
Updated the square toggle to match that in the new components area of IRB


tdumitrescu:
doesn't this need flat vs skeuo? or is square skeuo not going to be a thing

jayjo:
@tdumitrescu Nope, after our last design review we decided against having a skeuo square toggle and reverted back to semi-modern mixpanel tooltip

PR #166, opened 2017-02-12
update some styles and add close button to tutorial-tooltip

jbwyme:


lib/stylesheets/mixins/common.styl
```diff
@@ -8,9 +8,10 @@
 disabled = '&:disabled,&[disabled]'
 not-disabled = ':not([disabled]):not(:disabled)'
 
-z-modal = 9000
-z-overlay = 8000
-z-dropdown = 7000
+$z-modal = 9000
+$z-overlay = 8000
+$z-dropdown = 7000
+$z-tooltip = 6000
```
evnp: I love these

lib/widgets/tutorial-tooltip/index.js
```diff
@@ -15,6 +15,10 @@ registerMPElement(`mp-tutorial-tooltip`, class extends Component {
         delayRemove: 400, // TODO replace this with "mp-slow" from stylesheets/mixins/timing.json
       },
       helpers: {
+        close: e => {
+          e.stopPropagation();
```
evnp: total nit, can we standardize on ev or e? I don't care which, I believe we consistently use ev in irb but would be an easy update

jbwyme: I always use `e`, but I don't really care which we use. Feel free to pick one and send out an email to product eng haha.

PR #169, opened 2017-02-14
added a box-sizing declaration to mp-toast

jayjo:
The `min-width` of the toast widget is set to `354px`, however because there's no `box-sizing` property the toast is adding the padding to the overall container width making it ~50px wider than it should be.


lib/components/toast/index.styl
```diff
@@ -28,6 +28,7 @@
         background-color mp-blue
         border-radius 6px
         box-shadow 0 8px 22px 0 rgba(0, 0, 0, 0.37)
+        box-sizing border-box
```
zakj: I feel like we should just add this to everything by default rather than trying to patch it in to fix these issues that inevitably crop up. @tdumitrescu @jbwyme any reason we don't do that already?

jayjo: I agree - I think we'd have to make some fairly serious adjustments to things with fixed widths, though. Is there a way to have it touch only newly built things? Might be less of a headache to clean up.

jayjo: @jbwyme ^

jbwyme: Yeah our goal has been to add it to everything by default when creating new components (see https://github.com/mixpanel/mixpanel-common/blob/master/lib/components/button/index.styl#L28). I think we copied this component from a one-off and missed it.

jbwyme: We can't globally add it though because of the reasons you mentioned: it would conflict with a lot of other existing stuff.

PR #172, opened 2017-02-16
mp-truncated-text element to responsively truncate middle of text

jhp:


lib/util/index.js
```diff
@@ -181,6 +182,50 @@ export function truncateMiddle(string, len) {
   return string;
 }
 
+/*
```
jalack: nit: JSDoc style for comments?

lib/util/index.js
```diff
@@ -181,6 +182,50 @@ export function truncateMiddle(string, len) {
...
+export function truncateToElement(string, elem) {
+  var availableWidth = elem.clientWidth;
+  var styles = window.getComputedStyle(elem);
+  availableWidth -= parseInt(styles.getPropertyValue(`padding-left`));
+  availableWidth -= parseInt(styles.getPropertyValue(`padding-right`));
+  return truncateToWidth(string, styles.getPropertyValue(`font`), availableWidth);
+}
+
+
+
```
jalack: nit: remove extra lines

lib/util/index.js
```diff
@@ -181,6 +182,50 @@ export function truncateMiddle(string, len) {
...
+  if (fn(start) >= 0) {
+    return start;
+  }
+  if (fn(end) < 0) {
+    return end;
+  }
+  while(end - start > 1) {
+    var mid = Math.floor((end + start) / 2);
+    var fmid = fn(mid);
+    if(fmid < 0) {
```
jalack: nit: spaces after `if` 😄

lib/components/truncated-text/index.js
```diff
@@ -0,0 +1,51 @@
...
+
+import { registerMPElement } from '../../util/register-element.js';
+import { truncateToElement } from '../../util';
+
+registerMPElement(`mp-truncated-text`, class extends WebComponent {
+
+  updateText(force) {
+    var newWidth = this.clientWidth;
+    if(!force) {
+      // only update once in 150 ms.
```
jalack: shall we bring in/use a `throttle` for resize events instead of implementing this way?

lib/util/index.js
```diff
@@ -181,6 +182,50 @@ export function truncateMiddle(string, len) {
...
+  return end;
+}
+
+const measureContext = document.createElement(`canvas`).getContext(`2d`);
+
+export function truncateToWidth(string, font, width) {
+  measureContext.font = font;
+  const fitFn = len =>
+        width - measureContext.measureText(truncateMiddle(string, string.length - len)).width;
+  const bestFit = binarySearch(0, string.length - 4, fitFn);
```
jalack: why 4? for the ellipses? what about the case where we can fit the whole string?

jhp: it's a little confusing here but this parameter is being subtracted from the string length, so 0 corresponds to the whole string and string.length - 4 corresponds to 4 characters.

examples/style-guide/index.jade
```diff
@@ -39,6 +39,9 @@
     h2 Utility classes
     p.mp-font-paragraph .mp-font-paragraph
       p Lorem ipsum dolor sit amet, consectetur adipiscing elit. Aliquam dignissim sodales rhoncus. Cras rutrum nisl et sapien efficitur tempus. Sed sagittis sed lorem in efficitur. Morbi sollicitudin interdum sapien, vitae semper odio dapibus non. Aenean id massa id nunc cursus varius. Suspendisse congue leo ac ante egestas molestie. Phasellus nec arcu in dolor tincidunt ornare a sed nunc. Proin et dapibus augue, sit amet fermentum risus. Integer ornare convallis nulla, in condimentum lectus efficitur et. Fusce facilisis, ex vitae consequat scelerisque, enim nibh fringilla ligula, a bibendum orci lorem id felis. Mauris eget augue nibh. Ut quis pharetra velit. Donec tempor ante id risus porttitor venenatis. Proin id dictum arcu, pulvinar iaculis augue. Aliquam in nunc vitae urna feugiat condimentum.
+    p
+      responsive truncated text
+      mp-truncated-text(attrs={title: "browser text measurement is used to truncate this text so it can even handle strangely-sized characters like ﷽﷽﷽﷽﷽﷽. Text will be re-truncated to fit the actual width of the containing element"})
```
jhp: I think we're almost always wanting a title set, since it gives the full text on mouseover, so I thought I'd just use that directly.

lib/components/truncated-text/index.js
```diff
@@ -0,0 +1,51 @@
...
+    // in case the size of the element changes without the window
+    // resizing
+    this.interval = setInterval(this.listener, 500);
+    window.addEventListener(`resize`, this.listener);
+  }
+
+  detachedCallback() {
+    if(this.interval) {
+      clearInterval(this.interval);
+      this.interval = null;
```
jhp: If `detachedCallback` is somehow called twice without `attachedCallback` being called, which does seem to happen judging by exceptions in Panel, then we won't end up double-calling clearInterval or removeEventListener. I'm not sure if that would do any harm, but I think it's just better not to rely on behavior that's outside of normal parameters, like calling removeEventListener on a function that hasn't been set as an event listener.

I also just usually follow the pattern of removing event listeners or intervals after they've fired, it makes it easier to tell what might be in progress.

lib/components/truncated-text/index.js
```diff
@@ -0,0 +1,51 @@
...
+    this.updateText(true);
+  }
+
+  attachedCallback() {
+    this.updateText();
+    
+    this.listener = () => this.updateText();
+    // in case the size of the element changes without the window
+    // resizing
+    this.interval = setInterval(this.listener, 500);
```
jhp: Well this code does react to certain events: namely window resize and attribute update. The polling is a fallback for other resizing.

I took a look at that `element-resize-event` library but it makes me nervous how it's using all these `<object>` elements. `<object>` seems to be mostly used for flash and I worry that this use case for resize callbacks isn't really in line with what browser developers expect, so it might have bugs or performance issues. It also modifies the DOM and then relies on that DOM modification to persist, so this is a somewhat unexpected hidden requirement of the library and IMO a probable source of bugs (for example patching with a new VDOM could simply remove the injected `<object>`, or if you set `innerHTML` etc).

lib/components/truncated-text/index.js
jhp: I'd be happy to use `throttle`. Since `_` wasn't available in `mp-common` I thought I'd avoid adding it by just setting a timestamp here.

lib/components/truncated-text/index.js
jalack: word, sounds good to me

examples/style-guide/index.jade
jalack: I personally would rather require the user of the component to be more explicit and make it clearer what attributes are set, but I'm not too particular on this one

lib/components/truncated-text/index.js
jalack: fair objections to the library -- my concern is that we may want resizing for events that are not window resizing; for example, expanding a card when another one is deleted, for example. I suppose we could add some mechanism for having a client explicitly call for an update of this component, but ideally we'd like it to be self-contained

lib/components/truncated-text/index.js
jalack: as always, there's a [package](https://www.npmjs.com/package/lodash.throttle) for that

lib/util/index.js
```diff
@@ -181,6 +181,48 @@ export function truncateMiddle(string, len) {
...
+  return truncateMiddle(string, string.length - bestFit);
+}
+
+export function truncateToElement(string, elem) {
+  var availableWidth = elem.clientWidth;
+  var styles = window.getComputedStyle(elem);
```
jalack: nit nit: `let` and `const` respectively on the lines above

lib/util/index.js
```diff
@@ -181,6 +181,48 @@ export function truncateMiddle(string, len) {
...
+export function binarySearch(start, end, fn) {
+  if (fn(start) >= 0) {
+    return start;
+  }
+  if (fn(end) < 0) {
+    return end;
+  }
+  while (end - start > 1) {
+    var mid = Math.floor((end + start) / 2);
+    var fmid = fn(mid);
```
jalack: nit: unnecessary variable

lib/components/truncated-text/index.js
tdumitrescu: Elsewhere we install the main lodash package and import only the modules we need: `import throttle from 'lodash/throttle'`. This should lead to a more efficient bundle once you use multiple lodash functions that share dependencies.

lib/components/truncated-text/index.js
tdumitrescu: Here's my only concern with the current listener setup: I can imagine we'd want to use an element like this for long lists (menu items, table columns, filter results) - if we have 1500 of them on a page do we want 3000 callbacks/sec firing when idle? What about sharing a single module-scoped listener across all instances, which can then trigger resize calls on instances when an actual change happens?

examples/style-guide/index.jade
jbwyme: I'm ok with using title assuming that we don't get _both_ a browser tooltip and our custom tooltip when a tooltip is visible. It appears that we will though, is that correct?

lib/util/index.js
```diff
@@ -181,6 +181,47 @@ export function truncateMiddle(string, len) {
...
+
+export function truncateToWidth(string, font, width) {
+  measureContext.font = font;
+  const fitFn = len =>
+        width - measureContext.measureText(truncateMiddle(string, string.length - len)).width;
+  const bestFit = binarySearch(0, string.length - 4, fitFn);
+  return truncateMiddle(string, string.length - bestFit);
+}
+
+export function truncateToElement(string, elem) {
```
jbwyme: can we get some test coverage on these utils?

lib/util/index.js
jbwyme: #functionalprogramming sorry I'm slow: what's the specific usefulness of having 4 characters? In any case can we add some comments around here to explain what's happening. I've been looking at this method for about 10 minutes trying to reason through what exactly happens (slow like I said). I feel like this problem has been broken down into small pieces which is nice but it also makes the logic hard to follow through the various generic method calls.

lib/util/index.js
```diff
@@ -181,6 +181,47 @@ export function truncateMiddle(string, len) {
...
+  let availableWidth = elem.clientWidth;
+  const styles = window.getComputedStyle(elem);
```
jbwyme: do you have a sense of the performance hit we'll get calling this _n_ times every 0.5 seconds?

examples/style-guide/index.jade
jhp: I didn't realize we had a custom tooltip for these truncated texts, will switch to using 'text'.

jbwyme: no sorry i misunderstood, doing this is fine

lib/util/index.js
jhp: Hmm, looks like I misread the truncateMiddle function - I thought it only made a difference down to 4 chars, but actually it keeps shortening the string and just removes the ellipsis. So it would make sense to go down to 1.

lib/util/index.js
jhp: A resize test with 100 elements on Chrome shows 6ms for `measureText`, 3ms for `styles.getProperty()`.

PR #176, opened 2017-02-23
Date utils

evnp:
Move date utils from Insights here and add unit tests for them

corresponding Insights branch: https://github.com/mixpanel-platform/insights/pull/323

### DIFF
--- a/lib/util/date.js
+++ b/lib/util/date.js
@@ -53,85 +53,90 @@ const DISPLAY_DATE_FORMATS = {
 /**
  * Parse a date string into a Date object (sets date time to start of day)
  * Strings that parse to a future date (i.e. "12/31") are moved to the most recent *past* instance of that date
- * @param {String} dateString - string to parse
- * @param {Boolean} iso - treat input date string as an ISO-formatted UTC date (i.e. "2017-02-28")
- * @param {Boolean} startOfDay - optional; set date time to start of day
- * @param {Boolean} endOfDay - optional; set date time to end of day
- * @returns {Date|null} - the Date object the input string parses to, or null if input is invalid
+ * @param {string} dateString - string to parse
+ * @param {boolean} [iso] - treat input date string as an ISO-formatted UTC date (i.e. "2017-02-28")
+ * @param {boolean} [startOfDay] - set date time to start of day
+ * @param {boolean} [endOfDay] - set date time to end of day
+ * @returns {?Date} - the Date object the input string parses to, or null if input is invalid
  */
 export function parseDate(dateString, {iso=false, startOfDay=false, endOfDay=false}={}) {
-  if (typeof dateString === `string`) {
-    let date = null;
+  let date = null;
+  let mDate;
 
+  if (typeof dateString === `string`) {
     if (iso) {
-      date = moment.utc(new Date(dateString));
+      mDate = moment.utc(new Date(dateString));
     } else {
       // replace all chars other than alphanumerics and spaces with a space
       // this allows strings like '"2017-02-02"' to be parsed correctly
       dateString = dateString.replace(/[^\w\s]+/g, ` `);
-      date = moment(dateString, PARSE_DATE_FORMATS);
+      mDate = moment(dateString, PARSE_DATE_FORMATS);
     }
 
-    if (date.isValid()) {
-      date = startOfDay ? date.startOf(UNITS.day) : date;
-      date = endOfDay ? date.endOf(UNITS.day) : date;
+    if (mDate.isValid()) {
+      mDate = startOfDay ? mDate.startOf(UNITS.day) : mDate;
+      mDate = endOfDay ? mDate.endOf(UNITS.day) : mDate;
 
       // make dates like "12/31" parse to the most recent past instance of that date
-      date = date > moment() ? date.subtract(1, `${UNITS.year}s`) : date;
-
-      return new Date(date);
+      mDate = mDate > moment() ? mDate.subtract(1, `${UNITS.year}s`) : mDate;
+      date = new Date(mDate);
     }
   }
 
-  return null;
+  return date;
 }
 
 /**
  * Convert a Date object or integer timestamp into a formatted date string
- * @param {Date|Integer} date - Date object or timestamp to format
- * @param {Boolean} iso - optional; format date to ISO-style string (YYYY-MM-DD)
- * @param {Boolean} utc - optional; treat input as a UTC date
- * @param {String} unit - optional; use the corresponding format for this unit from DISPLAY_DATE_FORMATS
- * @param {Object} customFormatting - optional; map of units to custom format strings to be used in place of DISPLAY_DATE_FORMATS
- * @param {displayRangeIfWeek} - optional; if unit is week format to range string across week (i.e. "Dec 1 - Dec 7")
- * @returns {String|null} - the string the input Date object formats to, or null if input is invalid
+ * @param {date|number} date - Date object or timestamp to format
+ * @param {boolean} [iso] - format date to ISO-style string (YYYY-MM-DD)
+ * @param {boolean} [utc] - treat input as a UTC date
+ * @param {string} [unit] - use the corresponding format for this unit from DISPLAY_DATE_FORMATS
+ * @param {Object} [customFormatting] - map of units to custom format strings to be used in place of DISPLAY_DATE_FORMATS
+ * @param {boolean} [displayRangeIfWeek] - if unit is week format to range string across week (i.e. "Dec 1 - Dec 7")
+ * @returns {?string} - the string the input Date object formats to, or null if input is invalid
  */
 export function formatDate(date, {iso=false, utc=false, unit=null, customFormatting={}, displayRangeIfWeek=true}={}) {
+  let dateString = null;
+  let mDate;
+
   if (date instanceof Date || Number.isInteger(date)) {
-    date = utc ? moment.utc(Number(date)) : moment(Number(date));
+    mDate = utc ? moment.utc(Number(date)) : moment(Number(date));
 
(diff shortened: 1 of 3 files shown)


lib/util/date.js
```diff
@@ -0,0 +1,179 @@
...
+    .map(date => date > now ? now : date)
+    .map(date => date.getTime())
+    .sort((a, b) => a - b)
+    .map(date => formatDate(date, {iso: true}));
+}
+
+/**
+ * Returns a unit string corresponding with the time delta between two Dates
+ * @param {Date|Integer} from - the starting date of the range - can be Date or timestamp
+ * @param {Date|Integer} to - the ending date of the range - can be Date or timestamp
```
nojvek: `Date|Integer` this is just for human consumption right, it doesn't get validated does it?

test/util/date.js
```diff
@@ -0,0 +1,382 @@
+/* global describe, it */
```
nojvek: Great job on the tests. 

We should really track istanbul coverage for our utils in mixpanel-common.

test/util/date.js
```diff
@@ -0,0 +1,382 @@
...
+
+import {
+  parseDate,
+  formatDate,
+  normalizeDateStrings,
+  relativeToAbsoluteDate,
+  dateRangeToUnit,
+} from '../../lib/util/date';
+
+describe(`parseDate`, function() {
```
nojvek: BIKESHED: Should we use `()= > {` over `function() {`.  Less reading

lib/util/date.js
evnp: Just for human consumption yeah. I'd be happy to update the format if there's something typescript-compatible, was just following the conventions from other files

test/util/date.js
evnp: That'd be super cool. Might be worth trying to roll into my test infra updates this week, will revisit once I have a clearer idea of that

test/util/date.js
evnp: I don't have an issue with either. @tdumitrescu any binding-related reason not to do this? (once you're back)

tdumitrescu: [`mocha` wants control over `this`](https://mochajs.org/#arrow-functions).

lib/util/date.js
tdumitrescu: This is [jsdoc](http://usejsdoc.org/) format, with the point of eventually autogenerating docs.

evnp:
@tdumitrescu no rush on this but when you get a chance.. gonna wait for your signoff before merging

lib/util/date.js
```diff
@@ -0,0 +1,188 @@
...
+      date = endOfDay ? date.endOf(UNITS.day) : date;
+
+      // make dates like "12/31" parse to the most recent past instance of that date
+      date = date > moment() ? date.subtract(1, `${UNITS.year}s`) : date;
+
+      return new Date(date);
+    }
+  }
+
+  return null;
```
tdumitrescu: Pretty minor but I think these functions are cases where sticking to single-return will make following the control flow easier. Something along the lines of:
```js
let date = null;
if (typeof dateString === `string`) {
  // try to parse and set date
  if (momentDate.isValid()) {
    // more manipulation
    date = new Date(momentDate);
  }
}
return date;
```

lib/util/date.js
```diff
@@ -0,0 +1,188 @@
...
+}
+
+/**
+ * Convert a Date object or integer timestamp into a formatted date string
+ * @param {Date|Integer} date - Date object or timestamp to format
+ * @param {Boolean} iso - optional; format date to ISO-style string (YYYY-MM-DD)
```
tdumitrescu: the jsdoc notation for optional params is [brackets around the param name](http://usejsdoc.org/tags-param.html#optional-parameters-and-default-values):
```js
/**
 * @param {Boolean} [iso] - whether to format date to ISO-style string (YYYY-MM-DD)
 */
```

lib/util/date.js
```diff
@@ -0,0 +1,188 @@
...
+ * - sorts them in ascending order
+ * - formats them as ISO date strings (YYYY-MM-DD)
+ * @param {Array<String>} dates - list of date strings to normalize, can be any format that parseDate accepts
+ * @returns {Array<String|null>} - list of ISO-formatted normalized date strings, or null if the corresponding input date was invalid
+ */
+export function normalizeDateStrings(...dates) {
+  const now = new Date();
+  return dates
+    .map(parseDate)
+    .filter(date => date)
```
tdumitrescu: `.filter(Boolean)`?

lib/util/date.js
```diff
@@ -0,0 +1,188 @@
...
+  return null;
+}
+
+/**
+ * Accepts a list of date strings and
+ * - ensures none of them exceed the current moment
+ * - sorts them in ascending order
+ * - formats them as ISO date strings (YYYY-MM-DD)
+ * @param {Array<String>} dates - list of date strings to normalize, can be any format that parseDate accepts
+ * @returns {Array<String|null>} - list of ISO-formatted normalized date strings, or null if the corresponding input date was invalid
```
tdumitrescu: judging by github's jsdoc highlighting here, I'd guess this needs to be something like `@returns {Array.<String>|Array.<null>}` or `@returns {String[]|null[]}`.

lib/util/date.js
```diff
@@ -0,0 +1,188 @@
...
+
+/**
+ * Convert a Date object or integer timestamp into a formatted date string
+ * @param {Date|Integer} date - Date object or timestamp to format
+ * @param {Boolean} iso - optional; format date to ISO-style string (YYYY-MM-DD)
+ * @param {Boolean} utc - optional; treat input as a UTC date
+ * @param {String} unit - optional; use the corresponding format for this unit from DISPLAY_DATE_FORMATS
+ * @param {Object} customFormatting - optional; map of units to custom format strings to be used in place of DISPLAY_DATE_FORMATS
+ * @param {displayRangeIfWeek} - optional; if unit is week format to range string across week (i.e. "Dec 1 - Dec 7")
+ * @returns {String|null} - the string the input Date object formats to, or null if input is invalid
```
tdumitrescu: i think these nullable return types can be formatted like `@returns {?String}`; see http://usejsdoc.org/tags-type.html

test/util/date.js
```diff
@@ -0,0 +1,382 @@
...
+  it(`handles date strings of the desired formats`, function() {
+    const dec1st2015 = new Date(2015, 11, 1);
+    const feb27th2000 = new Date(2000, 1, 27);
+
+    ;[
```
tdumitrescu: why these semicolons? multiline array literals should work just fine without:
```js
[
  1,
  2,
  3,
].forEach(x => console.log(x));
```

that being said...
I'd prefer these to be named constants anyway.

test/util/date.js
```diff
@@ -0,0 +1,382 @@
...
+      `December 1 '15`,
+      `December 01 2015`,
+      `December 01 15`,
+      `December 01 '15`,
+      `December 1st 2015`,
+      `December 1st 15`,
+      `December 1st '15`,
+    ].forEach(dateString => expect(
+      parseDate(dateString).setHours(0, 0, 0, 0)
+    ).to.eql(dec1st2015.setHours(0, 0, 0, 0)));
```
tdumitrescu: how come you're redoing this `setHours` calculation every time? why not
```js
const DEC_1ST_2015 = new Date(2015, 11, 1).setHours(0, 0, 0, 0);
```

test/util/date.js
```diff
@@ -0,0 +1,382 @@
...
+      if (date > new Date()) {
+        date.setFullYear(date.getFullYear() - 1);
+      }
+
+      expect(parseDate(dateString))
+        .to.eql(date.setHours(0, 0, 0, 0));
+    });
+  });
+
+  it(`returns 'null' if invalid input is passed`, function() {
```
tdumitrescu: pretty nitpicky but i read this docstring thinking it returns the string "null"

test/util/date.js
```diff
@@ -0,0 +1,382 @@
...
+    ;[
+      ``,
+      `Ce n'est pas une date`,
```
tdumitrescu: Ceci

test/util/date.js
```diff
@@ -0,0 +1,382 @@
...
+  it(`returns null if invalid input is passed`, function() {
+    ;[
+      ``,
+      `Ce n'est pas une date`,
+      123,
+      null,
+      undefined,
+      [],
+      {},
+      new Date(),
```
tdumitrescu: you could extract the common parts of this repeated bad-input list into a shared constant

test/util/date.js
```diff
@@ -0,0 +1,382 @@
...
+    ].forEach(input => expect(relativeToAbsoluteDate(input, input)).to.eql(null));
```
tdumitrescu: maybe test when one argument is valid but the other isn't?

lib/util/date.js
evnp: i'm cool with it

lib/util/date.js
evnp: Nice, clearer

test/util/date.js
evnp: The linter actually hates them. I've just been bitten by annoying things like 
```
console.log('debug stuff')

[a, b] = vars;
```
which results in a very misleading error. Because of the linter issues I'll go with named constants as you say.

test/util/date.js
evnp: My thought was it makes it more clear that the comparison is disregarding time on both sides. I can see why you wouldn't want to sacrifice perf in here though, I'll make it clear in the var name.

test/util/date.js
evnp: touché

PR #185, opened 2017-03-03
always go through the _pendingAnimations flow to close or open.

### LINKED ISSUE #184: mp-confirm-delete stuck in weird state
https://mixpanel.github.io/mixpanel-common/examples/style-guide/index.html

Repro:
1. Double click the [Open confirm-delete] button real fast
2. the page will briefly start to get darker and then the modal will go away
3. click again -- the modal will not open again

If you inspect it after that, you can see it gets in a funny state where the `mp-modal-closed` class is on a lot of inner divs.
screen shot 2017-03-02 at 4 34 10 pm

crmorford:
This is because of an insights bug -- people were double-clicking the button that opens the delete modal, and so they were clicking the backdrop before the modal was fully open, which gets it stuck in a weird state.

fixes #184

### DIFF
--- a/lib/components/modal/index.js
+++ b/lib/components/modal/index.js
@@ -24,7 +24,7 @@ registerMPElement(`mp-modal`, class extends Component {
       },
       helpers: {
         backdropClicked: () => {
-          if (this.isAttributeEnabled(`closeable`) && this.state.visibility === VISIBILITY_OPEN) {
+          if (this.isAttributeEnabled(`closeable`)) {
             this.close();
           }
         },
--- a/lib/components/modal/index.js
+++ b/lib/components/modal/index.js
@@ -44,8 +44,6 @@ registerMPElement(`mp-modal`, class extends Component {
       case VISIBILITY_CLOSING:
         break;
       case VISIBILITY_OPENING:
-        this.update({visibility: VISIBILITY_CLOSED});
-        break;
       case VISIBILITY_OPEN:
         this._pendingAnimations = [`fadeModalOut`];
         if (this.config.helpers.getType() === `modal`) {
@@ -62,8 +60,6 @@ registerMPElement(`mp-modal`, class extends Component {
       case VISIBILITY_OPENING:
         break;
       case VISIBILITY_CLOSING:
-        this.update({visibility: VISIBILITY_OPEN});
-        break;
       case VISIBILITY_CLOSED:
         this._pendingAnimations = [`fadeModalIn`];
         if (this.config.helpers.getType() === `modal`) {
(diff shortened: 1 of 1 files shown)


lib/components/modal/index.js
```diff
@@ -24,7 +24,7 @@ registerMPElement(`mp-modal`, class extends Component {
       },
       helpers: {
         backdropClicked: () => {
-          if (this.isAttributeEnabled(`closeable`)) {
+          if (this.isAttributeEnabled(`closeable`) && this.state.visibility === VISIBILITY_OPEN) {
```
zakj: are you sure this is the right fix? the `close` method is attempting to do the right thing regardless of the `visibility` state; why isn't that switch working like it should?

tdumitrescu: the code of `close()` looks like it tries to handle non-open cases already, so this seems like a slightly misplaced bandaid. @jbwyme you know this `visibility` code better than anyone, opinions?

zakj: jinx

crmorford: aw darn it

jbwyme: I believe the problem you're running into is that it closes if the state is "OPENING": https://github.com/mixpanel/mixpanel-common/blob/master/lib/components/modal/index.js#L47

crmorford: bah i'm not sure. it just fixed my specific goal of "don't get into a weird state when the backdrop is clicked while opening", without making me afraid that i was gonna break something that every modal in all our apps (implicitly?) depends on by changing core logic 😨

jbwyme: ^ this is how code gets messy and things get broken in the long run

zakj:
can we kill off the _pendingAnimations stuff and use the snabbdom animation/delay features instead?

crmorford:
haha when i tried to look it up i found this: https://www.npmjs.com/package/snabbdom-delayed-class...do we really wanna use some code some weird stranger forked from snabbdom and put on the internet ;) ?

(but actually @zakj is this what you're talking about?)

tdumitrescu:
It wouldn't be a trivial rewrite, though we do have some other components using the newer pattern for self-animating, such as https://github.com/mixpanel/mixpanel-common/blob/master/lib/components/toast/index.jade. I'd involve @jbwyme if you do try it, since he did the original animation api for mp-modal.

crmorford:
Hm if it's not trivial i think hube wants me to stick to high-priority irb stuff for now :/
I'll come back after irb-clean-up-week

PR #186, opened 2017-03-03
we could just make them all 20px wider

crmorford:
just tryna nip these upsell-style complaints in the bud 
screen shot 2017-03-02 at 5 26 07 pm
screen shot 2017-03-02 at 5 25 34 pm

### DIFF
--- a/lib/widgets/upsell-popup/index.styl
+++ b/lib/widgets/upsell-popup/index.styl
@@ -1,7 +1,8 @@
 @import '../../stylesheets/defaults'
 
 mp-modal
-  width 285px
+  width 305px
+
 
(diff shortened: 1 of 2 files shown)


lib/widgets/upsell-popup/index.styl
```diff
@@ -26,11 +27,12 @@ mp-modal
...
-      .upgrade-link
-        color blue-900
-        cursor pointer
-        font-weight weight-bold
-        text-decoration none
 
         &:hover
           color mp-blue
+
+.upgrade-link
```
crmorford: this was getting the 'mp-upsell-popup' prefix & not being applied

PR #189, opened 2017-03-06
dashboard changes

ldelbeccaro:
- change color & hover color of mp-button-bar when grey (https://mixpanel.atlassian.net/browse/CA-2366)
- truncate middle in tag selector (https://mixpanel.atlassian.net/browse/CA-2345)
- open editTags by default when opening add to dash for existing report (https://mixpanel.atlassian.net/browse/CA-2263)


lib/widgets/add-to-dash/index.jade
```diff
@@ -45,7 +46,7 @@
             .mp-add-to-dash-loading
               .mp-add-to-dash-spinner
                 mp-spinner
-          else if editTags
+          else if editTags || existingReport
```
jbwyme: won't `existingReport` be true right after creating one since we set the bookmarks attribute as a part of saving?

ldelbeccaro: yeah that's a good point - I guess in that case there's no point for `reportCreated` to exist

jbwyme: i'm currently working on this component anyways so just ignore this problem for now i'll fix it

jbwyme: pull it out of this PR and i'll merge!

ldelbeccaro: done - you could check for !reportCreated and existingReport... maybe hacky though

PR #190, opened 2017-03-06
Refactor add-to-dash to support insights

jbwyme:


lib/widgets/add-to-dash/index.js
```diff
@@ -59,52 +57,48 @@ export default registerMPElement(`add-to-dash`, class extends Component {
...
           if (this.isAttributeEnabled(`show-tutorial-tooltip`)) {
             this.setAttribute(`show-tutorial-tooltip`, false);
           }
-          // add url hash to params to make it easy to navigate back to report from dashboard
-          const bookmark = this.getJSONAttribute(`bookmark`);
-          const type = this.getAttribute(`report-type`);
-          if (!bookmark) {
+
+          if (!this.isAttributeEnabled(`bookmark`)) {
+            // this implies we need to create a new bookmark
```
rickydoar: where do we actually create the new bookmark?

jbwyme: https://github.com/mixpanel/mixpanel-common/pull/190/files#diff-c82b3c469d6b947bace766bf0c48efe6R79

PR #193, opened 2017-03-08
lightning-bolt and lighting-bolt have the new [final final] shape 

crmorford:
lighting-bolt is dead, long live lighting-bolt

insights: https://github.com/mixpanel-platform/insights/pull/367


crmorford:
That was my plan!

Seems a lot safer -- and also I want the new bolt today and don't have time to QA the whole world.

PR #195, opened 2017-03-13
fix upsell-icon

jbwyme:


lib/widgets/add-to-dash/index.styl
```diff
@@ -13,21 +13,25 @@
...
+        right -6px
+        top -5px
 
     &.type-segmentation, &.type-retention
       .btn
         padding 14px
         svg-icon-style(grey-700, grey-900)
 
+        .upsell-icon
+          svg-icon-style(mp-purple, size: 16px)
```
jalack: not really related to this review, but something I've been wondering -- given that we use web components & prefixing, have we considered moving away from using hierarchical selectors & just making everything flat? we might need to tinker with our naming conventions a bit to do it, but would produce less verbose CSS and avoid any potential selector math nuisances

lib/widgets/add-to-dash/index.jade
```diff
@@ -31,13 +31,13 @@
...
-      if $component.isAttributeEnabled('over-free-limit') && !existingReport
-        .mp-add-to-dash-upsell-icon
-          svg-icon(attrs={icon: `lightning-bolt`})
       .mp-add-to-dash-icon
         svg-icon(
           attrs={icon: existingReport ? `dashboard-confirm` : `dashboard-add-to`},
         )
+        if $component.isAttributeEnabled('over-free-limit') && !existingReport
+          .mp-add-to-dash-upsell-icon
+            svg-icon(attrs={icon: `lighting-bolt`})
```
jalack: I only vaguely recall the great lightning v. lighting debate of 2017, but did you intend to use `lighting-bolt` here? and if so, those names are remarkably confusing

jbwyme: yes it's a clusterfuck but @crmorford is fixing that issue globally this week

lib/widgets/add-to-dash/index.styl
jbwyme: I'm not sure that anyone has put a lot of thought into it, not opposed but not necessarily in support of. In my experience, it's the exception that I have a selector specificity issue where I need to override something but have to do css gymnastics to do it. In this case, it's actually because `svg-icon-style` handles the hover color as well so we want the hit area of the main icon to be the whole button. However, there are two icons inside the button container (one for the button, one for the upsell). 

I'm not under the impression that the system is 100% efficient but I don't find I spend a lot of time on issues related to it so I'm inclined to just focus on lower hanging fruit.

PR #196, opened 2017-03-13
Remove stop propagation when removing tag

jalack:
here's the original change in which the `stopPropagation` was added: https://github.com/mixpanel/mixpanel-common/commit/fd2345d527bac6115baaf67d957110d8d591a4a2

doesn't seem like it's required any longer, and it interferes with listeners upstream. also found some inconsistent tests & behavior in the style guides so fixing those.


lib/components/tag/index.js
```diff
@@ -17,10 +17,7 @@ export default registerMPElement(`mp-tag`, class extends Component {
         sizeClass: `default-tag`,
       },
       helpers: {
-        remove: e => {
-          e.stopPropagation();
-          this.dispatchEvent(new CustomEvent(`change`, {detail: {action: `remove`}}));
-        },
+        remove: () => this.dispatchEvent(new CustomEvent(`change`, {detail: {action: `remove`}})),
```
jbwyme: so the issue i see is that this will break: https://github.com/mixpanel-platform/dashboard/blob/master/src/components/tag-filter/index.jade#L13 because clicking "close" will also fire a click event on the tag.

jalack: I see -- will create separate event for selection

lib/components/tag/index.js
```diff
@@ -17,9 +18,12 @@ export default registerMPElement(`mp-tag`, class extends Component {
         sizeClass: `default-tag`,
       },
       helpers: {
-        remove: e => {
-          e.stopPropagation();
-          this.dispatchEvent(new CustomEvent(`change`, {detail: {action: `remove`}}));
+        remove: () => this.dispatchEvent(new CustomEvent(`change`, {detail: {action: `remove`, tagName: this.getAttribute(`tag-name`)}})),
```
jbwyme: how come you decided not to use custom events here? by using something like `remove-tag` and `select-tag` we can drop all the stopPropagation stuff completely and don't have to use e.detail.

jalack: no good reason, just following out of date examples

PR #197, opened 2017-03-13
[mixpanel-common] Fix nestedObjectDepth to work for null values

Nicomoto:
Problem: 
typeof null => 'object' 
Object.keys(null) => Uncaught TypeError: Cannot convert undefined or null to object
Ugh.

Fixes rendering time series data when certain values are null (eg. avg of non existent values etc.)

### DIFF
--- a/lib/util/index.js
+++ b/lib/util/index.js
@@ -111,7 +111,7 @@ export function mapValues(obj, func) {
 }
 
 export function nestedObjectDepth(obj) {
-  return obj != null && typeof obj === `object` ? nestedObjectDepth(obj[Object.keys(obj)[0]]) + 1 : 0;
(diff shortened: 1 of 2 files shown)


lib/util/index.js
```diff
@@ -111,7 +111,7 @@ export function mapValues(obj, func) {
 }
 
 export function nestedObjectDepth(obj) {
-  return typeof obj === `object` ? nestedObjectDepth(obj[Object.keys(obj)[0]]) + 1 : 0;
+  return obj != null && typeof obj === `object` ? nestedObjectDepth(obj[Object.keys(obj)[0]]) + 1 : 0;
```
tdumitrescu: see linter error about preferring `!==` to `!=`

tdumitrescu:
published in 0.54.6

PR #199, opened 2017-03-15
Sorting improvements

evnp:
Adds
- ability to control individual orderings of numeric, date, base comparators within `numDateAlphaComparator`
- ability to pass a cache/map of date strings to timestamps to avoid parsing dates unnecessarily (optimization)

Tests in progress, will add docs once we're settled on the api/behavior

### DIFF
--- a/lib/util/array.js
+++ b/lib/util/array.js
@@ -106,11 +106,23 @@ export function baseComparator({order=`asc`, transform=identity}={}) {
  * <<< [`12~1`, `blarg`, `12~2`].sort(baseComparator({order: `desc`, transform: dateString => dateString.replace(`~`, ` `)}));
  * >>> [`12~2`, `12~1`, `blarg`] // sorted by parseable date value
  */
-export function dateStringComparator({order=`asc`, transform=identity, dateRegex=null, parseDateConfig={}}={}) {
+export function dateStringComparator({order=`asc`, transform=identity, dateRegex=null, parseDateCache=null, parseDateConfig={}}={}) {
+  dateRegex = dateRegex && new RegExp(dateRegex);
+
   return mapArguments(SORT_ORDERING[order], item => {
     item = transform(item);
-    const date = (!dateRegex || new RegExp(dateRegex).test(item)) && parseDate(item, parseDateConfig);
-    return date ? date.getTime() : null;
+    let timestamp = null;
+
+    if (!dateRegex || dateRegex.test(item)) {
+      if (parseDateCache) {
+        timestamp = parseDateCache[item] || null;
+      } else {
+        const date = parseDate(item, parseDateConfig);
+        timestamp = date && date.getTime();
+      }
+    }
+
+    return timestamp;
   });
 }
 
--- a/lib/util/array.js
+++ b/lib/util/array.js
@@ -1,3 +1,4 @@
+import { Cache } from './cache';
 import { parseDate } from './date';
 import {
   defaultOrdering,
@@ -80,40 +81,6 @@ const SORT_ORDERING = {
   desc: lexicalCompose(NULL_ORDERING, (a, b) => defaultOrdering(b, a)),
 };
 
-/** Cache object used to store sort comparison values to avoid redundant processing/parsing/etc. **/
-export class SortCache {
-  /**
-   * Create a SortCache
-   */
-  constructor() {
-    this._cache = {};
-  }
-
-  /**
-   * Add a value to the cache under key
-   * @param {string} key - the key to store the value under
-   * @param {*} value - the value to be stored
-   */
-  add(key, value) {
-    this._cache[key] = value;
-  }
-
-  /**
-   * Fetch a value from the cache under key.
-   * On cache miss, use defaultValue to return value (and store it in the cache).
-   * If no defaultValue supplied, cache miss will return null
-   * @param {string} key - the key to fetch a value for
-   * @param {*} [defaultValue] - if Function, will be called (with key as argument) to obtain value. Otherwise, will be used directly as value.
-   */
-  fetch(key, defaultValue=null) {
-    if (!this._cache.hasOwnProperty(key)) {
-      this._cache[key] = defaultValue instanceof Function ? defaultValue(key) : defaultValue;
-    }
-
-    return this._cache[key];
-  }
-}
-
 /**
  * Construct a sort comparator function that orders items by numeric value, comparing them via a <= operation.
  * null items are placed last in the sorted order.
@@ -133,7 +100,7 @@ export function baseComparator({order=`asc`, transform=identity}={}) {
  * then comparing via a <= operation. Invalid date strings are placed last in the sorted order.
  * @param {string} [order=`asc`] - `asc` (default) sorts ascending; `desc` sorts descending
  * @param {Function} [transform=identity] - a function that can transform each item before comparison (defaults to identity: x => x)
- * @param {SortCache} [parseDateCache] - a SortCache instance used to store date strings <-> timestamps to avoid redundant parsing
+ * @param {Cache} [parseDateCache] - a Cache instance used to store date strings <-> timestamps to avoid redundant parsing
  * @param {object} [parseDateConfig] - a config object that provides fine-grained control over how date strings are parsed (see parseDate)
  * @param {string|RegExp} [dateRegex] - a regex that can be used to only try and parse certain strings as dates (for performance)
  * @returns {Function} - a sort comparator func that can be passed to Array.sort (for any given a and b returns one of [1, 0, -1])
@@ -144,12 +111,12 @@ export function baseComparator({order=`asc`, transform=identity}={}) {
  * If you want to cache parsed date values across multiple sorts for performance, initialize your own parseDateCache externally
  * and pass in that single instance:
  *
- *   const parseDateCache = new SortCache();
+ *   const parseDateCache = new Cache();
  * ... later, within your sorting code ...
  *   array.sort(dateStringComparator({parseDateCache}));
  */
 export function dateStringComparator({order=`asc`, transform=identity, parseDateCache=null, parseDateConfig={}, dateRegex=null}={}) {
-  parseDateCache = parseDateCache || new SortCache();
+  parseDateCache = parseDateCache || new Cache();
   dateRegex = dateRegex && new RegExp(dateRegex);
 
   return mapArguments(SORT_ORDERING[order], item => {
@@ -170,7 +137,7 @@ export function dateStringComparator({order=`asc`, transform=identity, parseDate
  * then comparing via a <= operation. Invalid numeric strings are placed last in the sorted order.
  * @param {string} [order=`asc`] - `asc` (default) sorts ascending; `desc` sorts descending
  * @param {Function} [transform=identity] - a function that can transform each item before comparison (defaults to identity: x => x)
- * @param {SortCache} [parseNumberCache] - a SortCache instance used to store number-like values <-> numbers to avoid redundant parsing
+ * @param {Cache} [parseNumberCache] - a Cache instance used to store number-like values <-> numbers to avoid redundant parsing
  * @returns {Function} - a sort comparator func that can be passed to Array.sort (for any given a and b returns one of [1, 0, -1])
  * @example
  * <<< [`n:-9,000`, `n:blarg`, `n:42`].sort(baseComparator({order: `desc`, transform: numericString => numericString.split(:)[1]}));
@@ -179,12 +146,12 @@ export function dateStringComparator({order=`asc`, transform=identity, parseDate
  * If you want to cache parsed number values across multiple sorts for performance, initialize your own parseNumberCache externally
  * and pass in that single instance:
  *
- *   const parseNumberCache = new SortCache();
+ *   const parseNumberCache = new Cache();
  * ... later, within your sorting code ...
  *   array.sort(dateStringComparator({parseNumberCache}));
  */
 export function numericComparator({order=`asc`, transform=identity, parseNumberCache=null}={}) {
-  parseNumberCache = parseNumberCache || new SortCache();
+  parseNumberCache = parseNumberCache || new Cache();
 
   return mapArguments(SORT_ORDERING[order], item => {
     item = transform(item);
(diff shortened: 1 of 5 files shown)


lib/util/array.js
```diff
@@ -106,11 +106,23 @@ export function baseComparator({order=`asc`, transform=identity}={}) {
...
   return mapArguments(SORT_ORDERING[order], item => {
     item = transform(item);
-    const date = (!dateRegex || new RegExp(dateRegex).test(item)) && parseDate(item, parseDateConfig);
-    return date ? date.getTime() : null;
+    let timestamp = null;
+
+    if (!dateRegex || dateRegex.test(item)) {
+      if (parseDateCache) {
+        timestamp = parseDateCache[item] || null;
+      } else {
```
jordanmnunez: I think we should try to parsing the date if there is a `parseDateCache` but it doesn't have the item. what do you think?

lib/util/array.js
```diff
@@ -106,11 +106,23 @@ export function baseComparator({order=`asc`, transform=identity}={}) {
  * <<< [`12~1`, `blarg`, `12~2`].sort(baseComparator({order: `desc`, transform: dateString => dateString.replace(`~`, ` `)}));
  * >>> [`12~2`, `12~1`, `blarg`] // sorted by parseable date value
  */
-export function dateStringComparator({order=`asc`, transform=identity, dateRegex=null, parseDateConfig={}}={}) {
+export function dateStringComparator({order=`asc`, transform=identity, dateRegex=null, parseDateCache=null, parseDateConfig={}}={}) {
```
nojvek: Am I the only one who thinks the python `=` are weird in js. Its a conversation for another day though.

lib/util/array.js
```diff
@@ -143,11 +155,11 @@ export function numericComparator({order=`asc`, transform=identity}={}) {
  * <<< [`12/1`, `blarg`, `-9,000`, `12/2`, `42`].sort(baseComparator({order: `desc`}));
  * >>> [`42`, `-9,000`, `12/2`, `12/1`, `blarg`]
  */
-export function numDateAlphaComparator({order=`asc`, transform=identity, dateRegex=null, parseDateConfig={}}={}) {
+export function numDateAlphaComparator({order=`asc`, transform=identity, dateRegex=null, parseDateCache=null, parseDateConfig={}}={}) {
```
nojvek: May be have an example in comment for parseDateCache. 

As someone who is not familiar with code, I have no idea when to use it.

lib/util/array.js
```diff
@@ -143,11 +155,11 @@ export function numericComparator({order=`asc`, transform=identity}={}) {
...
   return lexicalCompose(
-    numericComparator({order, transform}),
-    dateStringComparator({order, transform, dateRegex, parseDateConfig}),
-    baseComparator({order, transform})
+    numericComparator({order: order.number || order, transform}),
+    dateStringComparator({order: order.date || order, transform, dateRegex, parseDateCache, parseDateConfig}),
+    baseComparator({order: order.base || order, transform})
```
tdumitrescu: yeah this definitely needs test and doc updates

lib/util/array.js
```diff
@@ -106,11 +106,23 @@ export function baseComparator({order=`asc`, transform=identity}={}) {
...
+  dateRegex = dateRegex && new RegExp(dateRegex);
+
   return mapArguments(SORT_ORDERING[order], item => {
     item = transform(item);
-    const date = (!dateRegex || new RegExp(dateRegex).test(item)) && parseDate(item, parseDateConfig);
-    return date ? date.getTime() : null;
+    let timestamp = null;
+
+    if (!dateRegex || dateRegex.test(item)) {
+      if (parseDateCache) {
```
tdumitrescu: need some tests around this `parseDateCache` behavior. also doc update. these utils are getting hairy.

evnp: definitely plan to add test and docs, I wanted feedback on the implentation first (see pr description)

lib/util/array.js
tdumitrescu: it might even be nice to pass in a smart cache that manages itself (which you could also instantiate in your calling code in Insights):
```js
parseDateCache = parseDateCache || new MyInMemoryCache();
timestamp = parseDateCache.fetch(item, function notFound(item) {
  const date = parseDate(item, parseDateConfig);
  return date && date.getTime();
});
```
but that may be getting a little fancier than we need right now

evnp: The thing is we know if a property value is a date in insights, it'll be in `parseDateCache`. So by trying to parse things that aren't in the cache, we're just unecessarily parsing strings that we already know aren't dates. Maybe it should be an option? Kinda wondering if I should just split this new logic off into a separate comparator, something like
```
cacheComparator({order=`asc`, transform=identity, cache={}}={})
```

lib/util/array.js
```diff
@@ -143,11 +155,11 @@ export function numericComparator({order=`asc`, transform=identity}={}) {
  * <<< [`12/1`, `blarg`, `-9,000`, `12/2`, `42`].sort(baseComparator({order: `desc`}));
  * >>> [`42`, `-9,000`, `12/2`, `12/1`, `blarg`]
  */
-export function numDateAlphaComparator({order=`asc`, transform=identity, dateRegex=null, parseDateConfig={}}={}) {
+export function numDateAlphaComparator({order=`asc`, transform=identity, dateRegex=null, parseDateCache=null, parseDateConfig={}}={}) {
   return lexicalCompose(
-    numericComparator({order, transform}),
-    dateStringComparator({order, transform, dateRegex, parseDateConfig}),
-    baseComparator({order, transform})
+    numericComparator({order: order.number || order, transform}),
```
evnp: good call, thanks

lib/util/array.js
tdumitrescu: aren't you already trying to parse things that aren't in the cache?

evnp: didn't refresh and see your comment ted, I like that. I was parsing everything previously but it seems like we don't need to try and parse as date in a lot of cases, as long as we can ensure all dates are in the cache. This would be good for performance, afaik moment parsing isn't the fastest thing in the world. I'm gonna try and clear this up

lib/util/array.js
```diff
@@ -80,6 +80,40 @@ const SORT_ORDERING = {
   desc: lexicalCompose(NULL_ORDERING, (a, b) => defaultOrdering(b, a)),
 };
 
+/** Cache object used to store sort comparison values to avoid redundant processing/parsing/etc. **/
+export class SortCache {
+  /**
+   * Create a SortCache
+   */
+  constructor() {
+    this._cache = {};
```
tdumitrescu: might it be better to use a Map than an object? Since the key values can come from user input, Map will be safer by avoiding the chance of overwriting object built-in properties. e.g.:
```js
myCache.fetch(`hasOwnProperty`, 5);
myCache.fetch(`any value`); // BOOM!
```
maybe even worth a unit test for that case

lib/util/array.js
```diff
@@ -80,6 +80,40 @@ const SORT_ORDERING = {
...
+  /**
+   * Fetch a value from the cache under key.
+   * On cache miss, use defaultValue to return value (and store it in the cache).
+   * If no defaultValue supplied, cache miss will return null
+   * @param {string} key - the key to fetch a value for
+   * @param {*} [defaultValue] - if Function, will be called (with key as argument) to obtain value. Otherwise, will be used directly as value.
+   */
+  fetch(key, defaultValue=null) {
+    if (!this._cache.hasOwnProperty(key)) {
+      this._cache[key] = defaultValue instanceof Function ? defaultValue(key) : defaultValue;
```
tdumitrescu: did you want to use `this.add()` here? or just delete `add()` altogether?

evnp: actually use `add` in insights where I pre-populate the cache, I'll use it here though (derp)

test/util/util.js
```diff
@@ -129,6 +133,15 @@ describe(`numericComparator()`, function() {
       transform: n => n % 3,
     }))).to.eql([5, 7, 3]);
   });
+
+  it.only(`caches values in parseNumberCache`, function() {
```
evnp: oops. Out of curiosity do you know of a cli way to run individual tests with Mocha? Not the first time I've accidentally committed this, googling wasn't too fruitful (`--grep` option kinda works but didn't seem very consistent iirc)

test/util/util.js
```diff
@@ -269,6 +291,111 @@ describe(`numDateAlphaComparator()`, function() {
...
+    expect([`12/31/16`, `January 8th, 2000`, `4-1-15`, `9.876`, `11`, `000`, `-12,000`, `efg`, `abc`, `xyz`].sort(numDateAlphaComparator({order: {
+      date: `desc`,
+      base: `desc`,
+    }}))).to.eql([`-12,000`, `000`, `9.876`, `11`, `12/31/16`, `4-1-15`, `January 8th, 2000`, `xyz`, `efg`, `abc`]);
+
+    expect([`12/31/16`, `January 8th, 2000`, `4-1-15`, `9.876`, `11`, `000`, `-12,000`, `efg`, `abc`, `xyz`].sort(numDateAlphaComparator({order: {
+      number: `desc`,
+      base: `desc`,
+    }}))).to.eql([`11`, `9.876`, `000`, `-12,000`, `January 8th, 2000`, `4-1-15`, `12/31/16`, `xyz`, `efg`, `abc`]);
+  });
```
evnp: I felt the complexity of these cases benefited from simply writing out the inputs and outputs inline, would be open to different organization if you prefer

lib/util/array.js
evnp: ahh good call

test/util/util.js
tdumitrescu: I've always just used `mocha -g`, never had an issue with it

test/util/util.js
tdumitrescu: No no, this is very straightfoward - we'd probably lose legibility and simplicity by DRYing out further. I might reformat these examples slightly to make the pre- and post-sort arrays more directly comparable:
```js
it(`defaults to "asc" order for unspecified independent sorts`, function() {
  let original = [`12/31/16`, `January 8th, 2000`, `4-1-15`, `9.876`, `11`, `000`, `-12,000`, `efg`, `abc`, `xyz`];
  let expected = [`11`, `9.876`, `000`, `-12,000`, `12/31/16`, `4-1-15`, `January 8th, 2000`, `abc`, `efg`, `xyz`];
  let comparatorOptions = {
    order: {
      date: `desc`,
      number: `desc`,
    },
  };
  expect(original.sort(numDateAlphaComparator(comparatorOptions))).to.eql(expected);
});
```
(note also the docstring rewritten to follow the "it [does this]" pattern rather than "it [hi I'm writing a grammatically unrelated sentence!]")

evnp: coolio, I like that

evnp:
@tdumitrescu we got some cache tests, lmk what you think

lib/util/cache.js
```diff
@@ -0,0 +1,61 @@
...
+
+  /**
+   * Fetch a value from the cache under key.
+   * On cache miss, use defaultValue to return value (and store it in the cache).
+   * If no defaultValue supplied, cache miss will return null.
+   * @param {string} key - the key to fetch a value for
+   * @param {*} [defaultValue] - if Function, will be called (with key as argument) to obtain value. Otherwise, will be used directly as value.
+   */
+  fetch(key, defaultValue=null) {
+    if (!this._cache.has(key)) {
```
tdumitrescu: `this.has(key)`?

lib/util/cache.js
```diff
@@ -0,0 +1,61 @@
...
+  }
+
+  /**
+   * Remove and return a value from the cache under key.
+   * On cache miss, use defaultValue to return value.
+   * If no defaultValue supplied, cache miss will return null.
+   * @param {string} key - the key to fetch a value for
+   * @param {*} [defaultValue] - if Function, will be called (with key as argument) to obtain value. Otherwise, will be used directly as value.
+   */
+  pop(key, defaultValue=null) {
```
tdumitrescu: where is this used? I'm used to `pop()` being used only in a stack/queue context. this is a little confusing to me but more to the point: if we don't have any consumer for this method then let's remove it until we actually need it.

evnp: I realized a cache/map implementation seemed incomplete without the ability to remove items (`pop` is similar to python dict's pop). Personally think it's worthwhile to add `has`+`remove` here while I'm working on it, instead of waiting till they're necessary then needing someone else to regain context. I would expect these methods as a consumer of this class. `pop` I definitely care less about, happy to remove it. 

Do you feel strongly that we shouldn't add these methods until they're used somewhere else?

tdumitrescu:
I'm ok with 'remove' since it's useful in your unit tests. Pop i'm not
seeing the value of right now, since a cache shouldn't be treated like a
dict. Extracting values directly from a cache should be reserved for
(hopefully internal) cache-management utils. If your consumer is doing a
lot of direct manipulation of cache contents then it's probably the wrong
data structure.

evnp:
Makes sense. I started with `remove` behaving like `pop` then realized they should be separate and `pop` named accordingly but just `remove` wfm for simplicity.

evnp:
@tdumitrescu feedback addressed, let me know if there's anything else (and publish when ready if you'd be so kind)

test/util/cache.js
```diff
@@ -0,0 +1,100 @@
...
+];
+
+describe(`Cache`, function() {
+  it(`can add any type of key/value`, function() {
+    const cache = new Cache();
+
+    for (const key of INPUTS) {
+      for (const value of INPUTS) {
+        cache.add(key, value);
+        expect(cache._cache.get(key)).to.eql(value);
```
tdumitrescu: can you do this without reaching into `cache._cache`? `expect(cache.fetch(key)).to.eql(value)`?

test/util/cache.js
```diff
@@ -0,0 +1,100 @@
...
+
+    for (const key of INPUTS) {
+      for (const value of INPUTS) {
+        cache.add(key, value);
+        expect(cache.fetch(key)).to.eql(value);
+        cache.remove(key, value);
+        expect(cache.fetch(key)).to.eql(null);
+      }
+    }
+  });
```
tdumitrescu: what i'm missing here is tests for what i'd consider the default usage of this class: `fetch()` with a fallback, including both static value and callback. i'd add a `describe('fetch')` block to go through at least these cases:
- it populates the cache with a static value
- it populates the cache with the result of a callback
- it returns a value that's already in the cache
- it does not rerun the callback for a value which is already in the cache

evnp: Ah totally, I'm dumb. Will add those

test/util/cache.js
```diff
@@ -0,0 +1,138 @@
...
+          cache.remove(key)
+          expect(cache.fetch(key)).to.eql(undefined);
+          expect(cache.fetch(key, defaultValue)).to.eql(defaultValue);
+          expect(cache.fetch(key)).to.eql(defaultValue);
+        }
+      }
+    }
+  });
+
+  it(`can fetch any type of key/value with a defaultValue function supplied`, function() {
```
tdumitrescu: I really would much rather see this split into individual test cases for the different aspects of the `fetch` method rather than a couple big grabbag cases. Remember that test cases also serve as documentation for someone coming to this code for the first time, and it's ok to have some repetition between test cases. It will really be easier to follow (and to read in CI output) with `describe('fetch()')` block containing separate `it`s.

tdumitrescu:
Published v0.56.0 whew!!

evnp:
image
thanks teddd

PR #200, opened 2017-03-15
include localizedDate for project offsets

jordanmnunez:


lib/util/date.js
```diff
@@ -137,16 +137,43 @@ export function formatDate(date, {iso=false, utc=false, unit=null, customFormatt
 }
 
 /**
+ * Returns the date provided in project localized time. If no date specificed use now.
+ * This was largely taken from analytics/media/js/utility.js
+ * @param {string} utcOffset - the starting date of the range - can be Date or timestamp
```
evnp: Is this outdated? Says `string` but also "can be Date or timestamp"?

lib/util/date.js
```diff
@@ -137,16 +137,43 @@ export function formatDate(date, {iso=false, utc=false, unit=null, customFormatt
...
+    date.getUTCFullYear(),
+    date.getUTCMonth(),
+    date.getUTCDate(),
+    date.getUTCHours(),
+    date.getUTCMinutes(),
+    date.getUTCSeconds(),
+    date.getUTCMilliseconds()
+  );
+  if (utcOffset) {
+    date.setMinutes(date.getMinutes() + utcOffset);
```
evnp: Will this work with a `Date` (as described in docstring)? appears to me like only an integer will do

lib/util/date.js
```diff
@@ -137,16 +137,43 @@ export function formatDate(date, {iso=false, utc=false, unit=null, customFormatt
...
 
 /**
+ * Returns the date provided in project localized time. If no date specificed use now.
+ * This was largely taken from analytics/media/js/utility.js
+ * @param {string} utcOffset - the starting date of the range - can be Date or timestamp
+ * @param {Date|object} date - a Date object to be offset. if this is not provided the function will create a new Date
+ * @returns {Date} - a Date in "project time" according to the utcOffset provided
+ */
+
+export function localizedDate(utcOffset, date) {
```
evnp: being super bike/nit/shed here but I'd really like `localizeDate(date, utcOffset)` for naming and arg consistency with `parseDate` and `formatDate`. Also if `utcOffset` is an integer it'd be really nice to see `localizeDate(date, {utcOffset=0}={})` so that callsites like`localizeDate(date, 700)` are a little more transparent (also seems more consistent with your `utcOffset` option in other utils)

lib/util/date.js
```diff
@@ -137,16 +137,43 @@ export function formatDate(date, {iso=false, utc=false, unit=null, customFormatt
...
+/**
  * Convert a "relative date" integer (i.e. 5 [days ago]) and a unit into an "absolute" Date object
  * @param {number} relativeDateInt - integer that (along with unit) represnts a time offfset from the current date
  * @param {string} unit - specifies the unit of time to multiply the relativeDateInt by to calculate the absolute date
  * @returns {?Date} - the Date object the given relative date integer and unit resolve to, or null if input is invalid
  */
-export function relativeToAbsoluteDate(relativeDateInt, unit) {
+export function relativeToAbsoluteDate(relativeDateInt, unit, {utcOffset=null}={}) {
   let date = null;
+  const now = Number.isInteger(utcOffset) ? moment(localizedDate(utcOffset)) : moment();
```
evnp: Why do this outside of the check below? seems unnecessary to do the date processing if the inputs are invalid.

lib/util/date.js
```diff
@@ -160,8 +187,8 @@ export function relativeToAbsoluteDate(relativeDateInt, unit) {
  * @param {Array.<string>} dates - list of date strings to normalize, can be any format that parseDate accepts
  * @returns {Array.<?string>} - list of ISO-formatted normalized date strings, or null if the corresponding input date was invalid
  */
-export function normalizeDateStrings(...dates) {
-  const now = new Date();
+export function normalizeDateStrings(dates=[], {utcOffset=null}={}) {
+  const now = Number.isInteger(utcOffset) ? localizedDate(utcOffset) : new Date();
```
evnp: So it looks like `utcOffset` has to be an integer, is this consistent throughout? Could you add @param docs above each function that uses it?

test/util/date.js
```diff
@@ -388,3 +398,13 @@ describe(`dateRangeToUnit`, function() {
...
   });
 });
+
+describe(`localizedDate`, function() {
+  it(`gives the current time when offsetting by the timezone you are in`, function() {
+    expect(localizedDate(-NOW.getTimezoneOffset(), NOW)).to.eql(NOW);
+  });
+
+  it(`always returns a date `, function() {
+    expect(localizedDate(null, null)).to.be.a(Date);
```
evnp: Could run through the `NON_DATE_INPUTS` list here, would also be great to see handling of every combination of date+offset being valid/invalid

evnp: kinda like this https://github.com/mixpanel/mixpanel-common/pull/200/files#diff-e73f25ec6406bd8f135903f51c25b69cL385

lib/util/date.js
jordanmnunez: woooops! this is just plain wrong :|

lib/util/date.js
jordanmnunez: you would think date went first. However offset is the only required param. I am going to make this an obj with optional params. calling `localizedDate()` would just give utc time.

jordanmnunez:
@evnp comments addressed!

PR #205, opened 2017-03-17
Add mp-calendar component

mduan:
image


lib/components/calendar/index.js
```diff
@@ -0,0 +1,272 @@
...
+      for (const type of [`start`, `end`, `in`]) {
+        if (Array.from(row.childNodes).some(node => node.classList.contains(`is-${type}range`))) {
+          row.classList.add(`has-${type}range`);
+        }
+      }
+    }
+  }
+}
+
+document.registerElement(`mp-calendar`, Calendar);
```
mduan: Use `registerMPElement`

mduan: Use `registerMPElement`

lib/components/calendar/index.js
```diff
@@ -0,0 +1,270 @@
...
+      this.removeAttribute(name);
+    }
+  }
+
+  get minDate() {
+    return this.getDateAttribute(`min-date`) || null;
+  }
+
+  get maxDate() {
+    return this.getDateAttribute(`max-date`) || new Date();
```
evnp: Would it be possible to set no max date with this setup? Looks like explicitly setting `'max-date': null` would default to `new Date()`. Think I'd go with
```
return this.hasAttribute(`max-date`) ? this.getDateAttribute(`max-date`) : new Date();
```

lib/components/calendar/index.js
```diff
@@ -0,0 +1,270 @@
...
+  get isRange() {
+    return this.isAttributeEnabled(`is-range`);
+  }
+
+  getDateAttribute(name) {
+    if (this.hasAttribute(name)) {
+      return parseDate(this.getAttribute(name));
+    } else {
+      return null;
+    }
```
evnp: This is great, reduces a lot of code duplication below. `parseDate` will return `null` if the passed date is invalid, think it'd be worth throwing an error in that case? Seems a bit more developer-friendly since that's likely a mistake (you'd have to explicitly check for `'date-attr': null` since I'm guessing we still want that to e valid).

lib/components/calendar/index.js
```diff
@@ -0,0 +1,270 @@
...
+  }
+
+  setDateAttribute(name, date) {
+    if (date) {
+      this.setAttribute(name, formatDate(date, {iso: true}));
```
evnp: Same here, think I'd make things explicit and throw an error if `formatDate` ever returns `null`

lib/components/calendar/index.js
```diff
@@ -0,0 +1,270 @@
...
+    }
+
+    if (!this.isRange) {
+      this.picker.setDate(this.selectedDate, /* preventOnSelect= */true);
+    } else {
+      this.picker.setStartRange(this.selectedFromDate);
+      this.picker.setEndRange(this.selectedToDate);
+    }
+
+    if (this.selectedFromDate && !this.selectedToDate) {
```
evnp: This isn't you, but I'd actually make this check
```
!(this.selectedFromDate && this.selectedToDate) && (this.selectedFromDate || this.selectedToDate)
```
(or `!this.selectedFromDate ^ !this.selectedToDate`) to make this less dependent on the specific behavior of the widget - from always selected first.

lib/util/date.js
```diff
@@ -180,6 +180,31 @@ export function relativeToAbsoluteDate(relativeDateInt, unit, {utcOffset=null}={
 }
 
 /**
+ * Accepts a [from, to] date range and the focused date ('to' or 'from') being edited
+ * @param {Array.<string>} - date range [from, to]
+ * @param {string} - one of <`from`, `to`>
```
evnp: Can we include param names here?
```
+ * @param {Array.<string>} dates - date range [from, to]
+ * @param {string} focusedDate - one of <`from`, `to`>
```

lib/util/date.js
```diff
@@ -180,6 +180,31 @@ export function relativeToAbsoluteDate(relativeDateInt, unit, {utcOffset=null}={
 }
 
 /**
+ * Accepts a [from, to] date range and the focused date ('to' or 'from') being edited
+ * @param {Array.<string>} - date range [from, to]
```
evnp: Since this always needs an array of 2 as input, how about just making it 
```
normalizeDateRange(from, to, focused)
```
so the structure of the required input is simpler?

lib/util/date.js
```diff
@@ -180,6 +180,31 @@ export function relativeToAbsoluteDate(relativeDateInt, unit, {utcOffset=null}={
 }
 
 /**
+ * Accepts a [from, to] date range and the focused date ('to' or 'from') being edited
```
evnp: This doesn't tell me what the function does, as a reader I'd like to see the intention summarized here

lib/util/date.js
```diff
@@ -180,6 +180,31 @@ export function relativeToAbsoluteDate(relativeDateInt, unit, {utcOffset=null}={
...
+    if (from > to) {
+      return [from, from];
+    }
+  } else if (focusedDate === `to`) {
+    if (to < from) {
+      return [to, to];
+    }
+  }
+
+  return [from, to];
```
evnp: Is the desired behavior here to set both from and to to the same date if to comes before from or vice versa? That seems inferior to the current Insights behavior of swapping from<->to if they're in the wrong order (Jeff and I worked on this, was intentional). The swap seems more useful to the user, I don't think I'd ever want to select a range and then see them both move to the same date

package.json
```diff
@@ -76,6 +76,7 @@
     "mixpanel-browser": "2.9.16",
     "moment": "^2.17.1",
     "panel": "0.10.0",
+    "pikaday": "git://github.com/dbushell/Pikaday.git#a1704adef06e985a448cd8ab313dee5a5787a228",
```
evnp: ooh. What are we getting out of this revision?

lib/components/calendar/index.styl
```diff
@@ -0,0 +1,182 @@
...
+        border-left 1px solid grey-200
+
+      table.pika-table
+        th, td
+          &:first-child
+            padding-left 18px
+
+          &:last-child
+            padding-right 18px
+
```
evnp: Thanks for cleaning this up, looks a lot better

lib/util/date.js
evnp: outdated, disregard

lib/util/date.js
```diff
@@ -180,6 +180,31 @@ export function relativeToAbsoluteDate(relativeDateInt, unit, {utcOffset=null}={
 }
 
 /**
+ * Accepts a [from, to] date range and the focused date ('to' or 'from') being edited
+ * @param {Array.<string>} dates - date range [from, to]
```
evnp: Since this always needs an array of 2 as input, how about just making it 
```
normalizeDateRange(from, to, focused)
```
so the structure of the required input is simpler?

lib/components/calendar/index.js
mduan: Neat. Never used xor operator in code before.

lib/util/date.js
mduan: Discussed with Jeff. This is the desired behavior.

package.json
mduan: It looks to be so that we get this fix: https://github.com/dbushell/Pikaday/commit/a1704adef06e985a448cd8ab313dee5a5787a228. 

The latest npm release 1.5.1 (https://github.com/dbushell/Pikaday/releases/tag/1.5.1) which is before that commit.

mduan:
@evnp Can you take another look. I've addressed your comments.

Within `mp-calendar`, I've also changed the code to use `string` dates (e.g. `2017-01-15`) rather than `Date` dates since the `string` format is more useful for what `mp-calendar` needs to do.

I've also added a bunch of wct tests.

examples/style-guide/index.js
```diff
@@ -69,6 +69,27 @@ document.registerElement(`style-guide`, class extends Component {
...
+        handleInsertComponent: el => {
+          const textEl = document.createElement(`div`);
+          textEl.textContent = this.helpers.getElementStr(el);
+          const parentEl = el.parentElement;
+          parentEl.insertBefore(textEl, parentEl.childNodes[0]);
+        },
+
+        getCalendarEventHandlers: () => {
+          return {
+            selectDate: e => console.log(e),
```
evnp: Debug code?

test/components/calendar.js
```diff
@@ -0,0 +1,264 @@
...
+      window.requestAnimationFrame(() => cb());
+    };
+  });
+
+  describe(`setting attributes`, function() {
+    it(`setting min-date`, function(done) {
+      this.resetCalendar({'min-date': `2017-01-05`}, () => {
+        this.navigateCalendar({year: 2017, month: 1});
+        const disabledDayEls = this.calendar.el.querySelectorAll(`td.is-disabled:not(.has-event) .pika-day`);
+        this.verifyDayEl(disabledDayEls[disabledDayEls.length - 1], {year: 2017, month: 1, day: 4});
```
mduan: Good idea. Done.

PR #208, opened 2017-03-23
Add to dash tracking

jalack:


lib/widgets/add-to-dash/index.js
```diff
@@ -174,12 +175,32 @@ export default registerMPElement(`add-to-dash`, class extends Component {
...
+
   _getDashboardApiUrl() {
     const host = this.getAttribute(`mixpanel-host`);
     const projectId = this.getAttribute(`project-id`);
     return `${host}/api/app/projects/${projectId}/dashboard-reports`;
   }
 
+  attachedCallback() {
+    super.attachedCallback(...arguments);
+    if (this.getAttribute(`tracking`)) {
```
jbwyme: so my initial idea was to allow for _n_ number of projects in a format like: `{[project-token]: 'event-prefix'}`. This seems to only allow for one _additional_ project. IMO we should either:

1) Allow for _n_ number of projects and don't default ANY project (in this case we always track to the main mixpanel project)
2) Allow for only one _additional_ project on top of the mixpanel project, don't support a custom prefix, and have an attribute like `tracking-token`.

My preference is for 1 but either would suffice.

jalack: I'm inclined to agree on 1 being more appropriate than _n_, will change to clarify

PR #209, opened 2017-03-27
[wip] Implement components for list items

mduan:


examples/style-guide/index.jade
```diff
@@ -122,6 +122,100 @@
...
+      // it as an attribute. Otherwise, part of component interface is from attribute, while other part is from
+      // from attributes.
+      .list-item-container
+        mp-list-item(attrs={
+          'is-upper-case': true,
+        }) Top Events
+
+      .list-item-container
+        mp-list-item(attrs={
+          'has-icon': true,
```
jbwyme: IMO no reason to have both `has-icon` and `icon`, could drop `has-icon` and use `isAttributeEnabled('icon')`

examples/style-guide/index.jade
```diff
@@ -122,6 +122,100 @@
...
+      // Implementing it by the caller is not only a lot of code duplication, but it also involves the caller being
+      // being aware of the semantics of how mp-list-item works (e.g. it needs to understand when an item is
+      // non-selectable, since a non-selectable item needs darker grey color for the non-highlight parts vs. a
+      // selectable list-item. Implementing highlighting as component attribute is weird unless the label is also
+      // an attribute? Maybe can support label/label-parts coming from attribute or as content:
+      // if labelParts
+      //   +matches(labelParts)
+      // else
+      //   content
+      .list-item-container
```
jbwyme: I'd suggest that `mp-list-item` allows for arbitrary children and the caller can just use the existing jade mixin for highlighting, the caller would definitely be the one to define what is selectable and not-selectable so no reason the caller can't determine when to call the `+matches` mixin vs not IMO. Then we can drop this `label` and `label-parts` stuff. There's not much of a code difference between `'label-parts': JSON.stringify(['', 'Top ', 'Ev', 'ents'])` and `+matches('Top', 'Top Events')`

If we want to, `mp-secondary-action-list-item` and `mp-secondary-action-list-item` could abstract that away behind the `label-parts` attribute but I'd leave it off of the base `mp-list-item` component.

examples/style-guide/index.jade
```diff
@@ -122,6 +122,100 @@
             mp-list-item there is text here
             mp-list-item there is text here
 
+  .section.list-items
+    h1 List item
+
+    h2 Basic
+    .list-item-group
+      .list-item-container
```
tdumitrescu: I still think these should always be demo'd in a dropdown, the way they will be in The Real World.

examples/style-guide/index.jade
```diff
@@ -122,6 +122,100 @@
...
+        mp-list-item Top Events
+
+      // How to support uppercase? Using attribute, class, or dependent on caller? Seems more consistent to have
+      // it as an attribute. Otherwise, part of component interface is from attribute, while other part is from
+      // from attributes.
```
tdumitrescu: it seems to me like the choice is between making the caller use an attr, like here, or making the caller specify a text-transform for the light dom content which will be injected into shadow dom (`mp-list-item {text-transform: uppercase}`). The latter feels more natural to me for such a simple style, but I'm ok with either.

examples/style-guide/index.jade
tdumitrescu: as long as your icon isn't named `false`, lulz...

examples/style-guide/index.jade
```diff
@@ -122,6 +122,100 @@
...
+          'icon': `star-top-events`,
+        }) Top Events
+
+      // Highlighting will be handled by component using attribute or by caller? If handled by caller, caller
+      // will need to re-implement some of the styling logic to override the styles that highlighting styles overrode.
+      // Implementing it by the caller is not only a lot of code duplication, but it also involves the caller being
+      // being aware of the semantics of how mp-list-item works (e.g. it needs to understand when an item is
+      // non-selectable, since a non-selectable item needs darker grey color for the non-highlight parts vs. a
+      // selectable list-item. Implementing highlighting as component attribute is weird unless the label is also
+      // an attribute? Maybe can support label/label-parts coming from attribute or as content:
```
tdumitrescu: yeah we definitely want the normal 'content' case available. my 1st thought at making a content-based api for highlighting would be something like:
```html
<mp-list-item>
  <span class="mp-match"></span>
  <span class="mp-no-match">Top</span>
  <span class="mp-match"> Ev</span>
  <span class="mp-no-match">ents</span>
</mp-list-item>
```
but is it the case that shadow dom v1 won't allow host styling on those?

examples/style-guide/index.jade
```diff
@@ -122,6 +122,100 @@
...
+          'icon': `star-top-events`,
+        }) Top Events
+
+    h2 Multi-line
+    .list-item-group
+      .list-item-container
+        mp-multi-line-list-item(attrs={
+          'secondary-action': `delete`,
+          'is-selected': true,
+          'label': `Top Events`,
```
tdumitrescu: this one i also presume we'd want a free-content api, at least for the main label

lib/components/list-item/index.js
```diff
@@ -11,6 +11,29 @@ export default registerMPElement(`mp-list-item`, class extends Component {
       css,
       template,
       useShadowDom: true,
+      defaultState: {
```
jbwyme: is there any reason to use internal state here vs just using the attributes? Keeping internal state is a good way to get your view out of sync with your app state -- we've tried this in the past and come to the conclusion that letting the application dictate _all_ of the state is the way to go. We have examples in mixpanel-common that keep state but IMO it was a mistake. Components should take state in through attributes and dispatch events when things change but not manage their own state.

examples/style-guide/index.jade
mduan: The styling for non-uppercase is:

```
font-size: 12px;
font-weight: 500;
```

While for uppercase it's:

```
font-size: 11px;
font-weight: 600;
text-transform: uppercase;
```

It's more than just a `text-transform`

examples/style-guide/index.jade
mduan: With v1 slots, this might work because `mp-match` and `mp-no-match` are top-level nodes of the slot they get inserted into. (I'll have to test it to see if it works). However, if you wrapped those `spans` in an outer div, it would no longer work:

```
<div>
  <span class="mp-match"></span>
  <span class="mp-no-match">Top</span>
  ...
</div>
```

If we wanted to create a mixin for the `+matches` (like Josh suggested below), it would no longer work because a jade mixin must return a single top-level node (correct me if I'm wrong about this).

Also, I personally feel there's more tight-coupling with this content-based highlighting approach. The interface that needs to be synched between callers is the structural requirement that it be a sequence of span tags and the exact class names `mp-match` and `mp-no-match`. With an attribute, the interface is attribute name `label-parts` and knowing the format of the attribute value (a JSONified list)

examples/style-guide/index.jade
mduan: My response to Ted above addresses the limitation of using the content-based highlighting using mixin.

examples/style-guide/index.jade
mduan: Having free-content api for the main label makes sense here to be consistent with the fact that `mp-list-item` supports free-content api. Having the secondary label be free-content might make sense as well, although I'll need to think a bit more about if there are limitations to this.

lib/components/list-item/index.js
mduan: The state kept here is not really app state. It's just a reflection of the attribute values with possible sanitization/processing applied first. Allowing for sanitization/processing is one benefit of storing the attribute value is `this.state`. Another reason is that I want to call `this.update(...)` when any attribute changes to trigger a re-render. Alternatively, I could have just had a single `this.update()`, without actually specifying any state to update) in `attributeChangedCallback`.

tdumitrescu: for the case of this component, `attributeChangedCallback` [will already trigger an update](https://github.com/mixpanel/panel/blob/85e668ed2ed8a2da93ec28af2c005e5917c922ee/lib/component.js#L283), no need to do it yourself.

examples/style-guide/index.jade
mduan: I actually prefer not having it in a mp-drop-menu since just laying it out on the page allows more easily seeing all the possible configurations of mp-list-item at a glance. It also gives more space for displaying more information about these configurations. For example, I currently display the exact attributes necessary to get the displayed configuration:

image

I don't see much benefit to having them in a mp-drop-menu, since it's pretty to visualize what they will look like in a mp-drop-menu (since being inside a mp-drop-menu doesn't affect the appearance of the mp-list-item).

However, I'm not that opinionated either way, so I can put them inside a dropdown if that's your preference.

lib/components/list-item/index.js
mduan: Another benefit with the approach I have is that it makes easy to see all the attributes a component supports at a glance by just looking in `attributeChangedCallback` to see all the attributes the component supports. (Although this could be provided in other ways as well.)

examples/style-guide/index.jade
mduan: Given that you're not able to style nested elements in a slot, it seems to me that the intention of slots is to allows the caller to specify content that 1) should solely be styled by the caller, 2) should not depend on component state (e.g. if the component is being hovered). The component's role should only be to position the slot within the component.

I think even having the label be a slot already pushing the limits of what slots are meant for (since the label's color depends on if the component if being hovered). Having something be a slot and having the component style the slot adds mental overhead when thinking about whether a slot would work, because we need to now think about whether the component can style deep enough into the slot by just styling the root elements of the slot. Currently it works for the label, but say later on for some reason, we need to add a container div to the label slot, then the component can no longer style the slot.

mduan: Actually, thinking some more, I think the jade mixin `+matches` would work in this case because you can have the root `div` it returns be a named slot:

mixin matches(...)
  <div slot="matches">
    <span class="mp-match"></span>
    <span class="mp-no-match">Top</span>
    ...
  </div>
```

However, I still feel this kind of interface is brittle for the reasons mentioned above.

File: examples/style-guide/index.jade (hunk 3 above)
Comment (tdumitrescu) [inline examples/style-guide/index.jade]: I think being in the menu does affect how it's visualized; I'd rather this be realistic than extra convenient.

File: examples/style-guide/index.jade (hunk 5 above)
Comment (tdumitrescu) [inline examples/style-guide/index.jade]: FWIW a mixin does not have to return a single root node; it gets compiled to literally a JS function that gets called where it's used. here's the highlighter mixin we use elsewhere:
```jade
mixin highlight-matches(matches)
  each substr, si in matches
    span.substr(class={'matches-search': !(si % 2)})= substr
```

PR #223, opened 2017-04-18
Increase box-shadow for popup modal

mduan:
Talked with Eric and agreed to darken the drop shadow for popups.


mduan:
Already merged into master.

PR #224, opened 2017-04-19
pass not-fullscreen down to modal

crmorford:


lib/components/alert/index.jade
```diff
@@ -1,9 +1,9 @@
 - const proxyAttrs = {'modal-type': 'popup'};
-- ['closeable', 'open'].forEach(attr => proxyAttrs[attr] = $component.getAttribute(attr));
+- ['closeable', 'not-fullscreen', 'open'].forEach(attr => proxyAttrs[attr] = $component.getAttribute(attr));
 mp-modal(attrs=proxyAttrs)
   .mp-alert-container
     .mp-alert-body
-      .mp-alert-icon
+      .mp-alert-icon(class={'large-icon': $component.getAttribute('large-icon')})
```
jbwyme: would `isAttributeEnabled` be more appropriate here?

crmorford: yep 👍

PR #226, opened 2017-04-25
add edit-input component

jbwyme:
http://cometapp.us/c/Z9SaVEyD

### DIFF
--- a/lib/components/edit-input/index.js
+++ b/lib/components/edit-input/index.js
@@ -1,0 +1,159 @@
+import { Component } from 'panel';
+
+import { registerMPElement } from '../../util/register-element.js';
+
+import template from './index.jade';
+import css from './index.styl';
+
+const GLOBAL_MIN_WIDTH = 90;
+const SIZER_PROPS = [
+  `border`,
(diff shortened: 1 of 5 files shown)


lib/components/edit-input/index.js
```diff
@@ -0,0 +1,159 @@
...
+  `fontFamily`,
+  `fontSize`,
+  `fontWeight`,
+  `height`,
+  `letterSpacing`,
+  `padding`,
+  `whiteSpace`,
+];
+
+export default registerMPElement(`mp-edit-input`, class extends Component {
```
mduan: Could we separate `resize-input` from `mp-edit-input` and have `mp-edit-input` use `resize-input`? The filters in insights and signal also use the `resize-input`, so it'd be good to have the resizable input logic shareable.

lib/components/edit-input/index.styl
```diff
@@ -0,0 +1,70 @@
+@import '../../stylesheets/defaults'
+
+placeholder()
```
zakj: autoprefixer handles this

PR #230, opened 2017-05-01
Implement mp-items-menu for events/properties menus

mduan:
The API for the events/properties menu looks like this:

```javascript
mp-items-menu(
  attrs={
    'search-filter': ``,
    'empty-message': ``,
  }
  props={
    sections: [{
      label: ``,
      isLoading: <true, false>,
      items: [{
        hasPropertiesPill: <true, false>,
        icon: ``
        label: ``
        isSelected: ``,
        isDisabled: ``,
        ...
      }],
    }],
  }
)
```


jbwyme:
Yeah overall that makes sense. I'd vote to go ahead and remove the extra layer of nesting like you suggested.

jbwyme:
oh btw for props let's make them hyphenated (i forget what that case is called): `item-type`

lib/components/items-menu/index.js
```diff
@@ -0,0 +1,279 @@
...
+          return filteredSections.map(section => {
+            const items = section.items.filter(item => item.index < this.state.progressiveListSize);
+            return extend(section, {items});
+          }).filter(section => section.items.length > 0 || section.isLoading);
+        },
+
+        unwrapItem: item => this.unwrapItem(item),
+
+        handlePostPatch: () => {
+          this.dispatchEvent(new CustomEvent(`draw`));
```
mduan: I added a comment here to make it clearer:

>When this component is patched, the height of it could change. Send a draw event so users of the component can respond to the height change accordingly.

examples/style-guide/index.jade
```diff
@@ -145,6 +145,31 @@
...
+      )
+      if open.itemsMenu
+        .drop-menu-container
+          mp-drop-menu(
+            attrs={open: true}
+            on={change: e => $helpers.handleModalChange('itemMenu', e.detail.state)}
+          )
+            mp-items-menu(
+              attrs={'search-filter': itemsMenuSearchFilter},
+              props={sections: $helpers.getItemsMenuSections()}
```
mduan: I still prefer having it as a prop. Having it as an attr adds the overhead of serializing/deserializing a potentially large list of items. Even though it's a prop, it still supports the caller updating the value of `sections` since I've made `sections` a setter on the component.

tdumitrescu: Here's my reply to @jordanmnunez on this issue from an old Insights PR where I converted chart data from an attr to a prop:

*so is this factual: props on elements actually live as objects vs attrs must be strings.*

Props don't have to be objects. They're just literally JS properties (or "instance variables" if you're pretending they're class instances) on the JS representation of the HTML element (which in our case here is a Panel component/app).

re: attrs, it's possible that since snabbdom uses `setAttribute` we could pass non-strs through unchanged. I haven't even tried, because that's an implementation detail that we shouldn't depend on. Attributes get set traditionally via markup `<my-el attr=foo>` where they can only be strings.

*are there any downside to setting other state info as props? (besides a few extra methods on the comp)
are there any downsides in general?*

Yes.
- Settings props is an imperative API (`myEl.foo = 'bar'`, requires JS) as opposed to declarative (`<my-el foo=bar>`, just DTRT). Now, because of virtual-jade and snabbdom, we actually _do_ have a way to set props declaratively based on app state (`props={foo: 'bar'}`), but for generic components that aren't necessarily embedded within a Panel app that's not usually a possibility.
- passing around objects by reference complicates the dom-diffing picture significantly, meaning that we have to manage change-notification to some extent in our app code, or at least worry about it. let's take an example: I have an object `myConfigStuff = {foo: 'bar'}` which gets passed to an independent component (not linked to my main app via state). If it's serialized as an attribute, then the main app's dom-differ looks at it as a string every time and says "has the value of this attr changed? if so, call `setAttribute()` on the child. if not, do nothing." If however it's a prop, and you just change `myConfigStuff.foo = 'bla'` and then update, what happens? The dom-differ says "has the value of this prop changed?" The answer depends on whether the virtual dom lib does an expensive deep-clone+deep-equals or just a reference equality check. In the case of snabbdom, [it's the latter](https://github.com/snabbdom/snabbdom/blob/12abb54e/src/modules/props.ts#L22) and nothing will happen (because it's the same object before and after). In the case of our chart datasets, there's a bunch of processing that happens during render which ensures that it's going to be a new object every time the chart template gets rendered (which luckily isn't on every update), and so `set chartData()` will get called. I think that the heavy cost of serializing+deserializing+comparing big datasets makes it worth the tradeoff in this case, but it's _much_ easier for bugs to creep in ("why didn't my chart change?") and if you never want to worry about that, just use attrs and let simple string-comparison do everything.

TLDR: using props with objects or other not-easily-diffed types is hard to get right with dom-diffing, and we should generally use it just where optimization is necessary and we're willing to trade off simplicity for it.

NB: using an immutable data library could solve some of this pain - at the expense of making some of our write/update operations a lot heavier. We should profile carefully when investigating.

examples/style-guide/index.js
```diff
@@ -27,11 +28,13 @@ document.registerElement(`style-guide`, class extends Component {
           confirm: false,
           confirmDelete: false,
           menu: false,
+          itemMenu: false,
```
zakj: should be `itemsMenu` (plural)?

mduan: 👍  fixed.

zakj:
published as 0.74.0

lib/widgets/items-menu/index.js
```diff
@@ -0,0 +1,247 @@
...
+  TAB: 9,
+  ENTER: 13,
+  UP_ARROW: 38,
+  DOWN_ARROW: 40,
+};
+
+ItemsMenu.ITEM_TYPE_EVENT = ItemsMenu.prototype.ITEM_TYPE_EVENT = `event`;
+ItemsMenu.ITEM_TYPE_PROPERTY = ItemsMenu.prototype.ITEM_TYPE_PROPERTY = `property`;
+
+document.registerElement(`mp-items-menu`, ItemsMenu);
```
tdumitrescu: whoops `Failed to execute 'registerElement' on 'Document': Registration failed for type 'mp-items-menu'. A type with that name is already registered.`

you need to follow the pattern in the other components+widgets (i.e. [use `registerMPElement()`](https://github.com/mixpanel/mixpanel-common/blob/c295e9f5674fff9feb1e9919c9fdbc5898bf49b6/lib/widgets/add-to-dash/index.js#L13)) to avoid this error when multiple scripts include this component

PR #235, opened 2017-05-05
Helper function for animating screen transition

mduan:


mduan:
Will keep in signal repo for now

PR #238, opened 2017-05-06
Add switch

jalack:
Add an unlabeled switch for use in Journeys & Lexicon.

screen shot 2017-05-05 at 5 09 19 pm


lib/components/switch/index.js
```diff
@@ -0,0 +1,43 @@
...
+
+export default registerMPElement(`mp-switch`, class extends Component {
+  get config() {
+    return {
+      css,
+      template,
+      useShadowDom: true,
+      defaultState: {
+        on: false,
+        clickable: true,
```
zakj: any reason not to use the html-standard `disabled` attribute for disabling?

jalack: I'm not averse to that: `readonly` might also make sense. The only potential confusion I could see is that disabling an element often results in a visual change, which is not really the case here (outside of changing the cursor).

jalack: after reading the HTML spec, I think `disabled` does indeed make sense. making the change

zakj: VINDICATION

.gitignore
```diff
@@ -1,4 +1,5 @@
 .DS_Store
+.vscode
```
crmorford: this supposed to be here?

jalack: yeah I don't think we need people's individual workspace settings here. is there a good use case you can see for leaving as is?

zakj: I lean towards using [per-user global gitignores](http://stackoverflow.com/questions/7335420/global-git-ignore#7335487) for editor-specific stuff, otherwise every repository's gitignore is a graveyard of editor files depending on who's worked on the project

tdumitrescu: except for fricking .DS_Store

test/components/switch.js
```diff
@@ -0,0 +1,47 @@
+/* global beforeEach, describe, it */
```
zakj: probably just following a pattern here but for the record `/* eslint-env jasmine, mocha */` is a bit more future-proof

zakj:
lgtm, fix my two nits or don't at your preference, but please do rebase and resolve conflicts

test/components/switch.js
jalack: sweet! I was indeed being a copy-paste bot

PR #239, opened 2017-05-08
add "change" event for pre-snabb world

crmorford:


lib/widgets/tutorial-tooltip/index.js
```diff
@@ -21,6 +21,8 @@ export default registerMPElement(`mp-tutorial-tooltip`, class extends Component
         },
         next: e => {
           e.stopPropagation();
+          //TODO remove change event once notifictions 3.0 is upgraded to panel 0.10
+          this.dispatchEvent(new CustomEvent(`change`, { detail: { event: `next` } }));
```
tdumitrescu: Don't use `event` as a key, let's use `action`

examples/style-guide/index.jade
```diff
@@ -495,7 +495,7 @@
             'next-button': true,
           }
           on={
-            next: $helpers.incrementTooltipStep,
+            change: ev => ev.detail && ev.detail.event==='next' && $helpers.incrementTooltipStep(),
```
tdumitrescu: This is just the tmp style guide so it doesn't matter, but in general try to avoid inlining helpers into the template. The exception to this is when you want to close over some value (e.g. call a helper with the right list item as argument), but that's not the case here.

Also personally I'd still use the non-deprecated event here in this code, but no biggie.

tdumitrescu:
published 0.72.0 (with linter changes too)

PR #244, opened 2017-05-09
make select components work without snabb

crmorford:
this branch is poorly named because i added more things we need to make notif 3.0 work

* add one font-size i forgot to style
* and add change events to where we fire a bunch of custom events, since pre-snabb can't listen to custom events


lib/components/toast/index.js
```diff
@@ -16,13 +16,21 @@ export default registerMPElement(`mp-toast`, class extends Component {
...
   }
 
+  _triggerAction() {
+    //can remove 'change' event when notifications 3 updates to panel 0.10
+    this.dispatchEvent(new CustomEvent(`change`, {detail: {action: `action`}}));
+    this.dispatchEvent(new CustomEvent(`action`));
+  }
+
   _triggerClose() {
+    //can remove 'change' event when notifications 3 updates to panel 0.10
```
jordanmnunez: lets mark these as traditional todo:
`// TODO: remove 'change' event....`

lib/widgets/tutorial-tooltip/index.js
```diff
@@ -17,6 +17,7 @@ export default registerMPElement(`mp-tutorial-tooltip`, class extends Component
       helpers: {
         close: e => {
           e.stopPropagation();
+          this.dispatchEvent(new CustomEvent(`change`, {detail: {action: `close`}}));
```
jordanmnunez: add todo here too?

crmorford: oh good call, forgot about the old commit

PR #246, opened 2017-05-09
Pretrack

yarneo:
Added new icon for pretrack and added new renamings for properties sent with pretrack


zakj:
shipped in 0.73.0

PR #251, opened 2017-05-12
added fix for vertical centering in mp-alert modal

EricBenjamin:


EricBenjamin:
@tdumitrescu this small styling thing was causing some fuss and confusion. I think this is the fix.

EricBenjamin:
Still having issue with mp-modal-close-btn positioning needing to be centered here. Cannot target from within alert container.. bummer.

PR #253, opened 2017-05-12
Add mp-items-menu to new style guide

mduan:
Basically copied over from the old style guide.


EricBenjamin:
oops, did not mean to do that ^

PR #257, opened 2017-05-17
added compiled style guide to gitignore

antialias:


tdumitrescu:
I'm afraid this will break http://mixpanel.github.io/mixpanel-common, which is served via GitHub pages out of the repo.

antialias:
@tdumitrescu what if the style guide were published separately to github pages?https://github.com/mixpanel/mixpanel-common/pull/258

PR #258, opened 2017-05-17
script for publishing styleguide to gh-pages branch

antialias:


scripts/publish-styleguide.js
```diff
@@ -0,0 +1,6 @@
+#!/usr/bin/env node
+const ghPages=require('gh-pages');
+const execSync=require('child_process').execSync;
+
+execSync('npm build')
```
tdumitrescu: `npm run build`?

antialias: d'oh

antialias:
> the annoyance of the multi-branch workflow

Good to have some perspective on this. I suspect that if the multi-branch workflow was annoying, then is was a bit more manual than running `npm run publish-styleguide` to publish the styleguide from from your current branch.

My motivation here was to get rid of the build process changing tracked files in git; it makes my diffs really noisy whenever I build mixpanel-common locally.

tdumitrescu:
That's weird - the `build` task shouldn't be modifying anything that's checked in. `build-examples` does, and you're only supposed to run it when needing to publish style guide. (In dev mode, webpack dev server will build style guide without modifying the on-disk compiled file.)

antialias:
> That's weird - the `build` task shouldn't be modifying anything that's checked in.

I was actually running `npm run watch`

tdumitrescu:
Maybe we should get rid of or rename that task, i didn't know anyone used it anymore. `npm start` will build the same files without writing style guide to disk. And all the extra benefits of webpack dev server. Are you ok with using that instead?

antialias:
`npm start` doesn't write the bundles to disk. I'm using mixpanel-common as a [link](https://yarnpkg.com/lang/en/docs/cli/link/)ed dependency of [`analytics`](https://github.com/mixpanel/analytics)

tdumitrescu:
Ah is that what you're trying to do. Then how about we just replace the contents of the current `watch` script with `node build.js && node build.js -w`?

antialias:
Sure, that would be fine with me as a quick solution.

Re: this pull request, I'd still like it to be considered; It's wise to avoid tracking a build in `master`, and I think the `gh-pages` module invoked via `npm run publish-styleguide` removes the pain points that you mentioned from the prior publishing workflow.

antialias:
bump @jbwyme

jbwyme:
assuming we aren't building the style-guide stuff by default anymore (e.g. `npm run watch`), what reason do we have for this branch? The command we use to publish new version of mixpanel-common (`mix release mpcommon`) already builds the style-guides (https://github.com/mixpanel/analytics/blob/master/mixfile/release.py#L49) so master should always reflect the latest code with no additional work. Is that not the desired effect?

PR #260, opened 2017-05-18
Make mp-modal close lighter on hover

zachwe:
https://mixpanel.atlassian.net/browse/PA-1249


jayjo:
👍

zakj:
published as 0.75.3

PR #261, opened 2017-05-18
watch no longer builds styleguide

antialias:


package.json
```diff
@@ -8,7 +8,8 @@
     "build-examples": "npm run build && webpack --context examples --config examples/webpack.config.js",
     "build-test": "webpack --context test --config test/webpack.config.js",
     "deploy-design": "./deploy-design.sh",
-    "watch": "npm run build && concurrently 'node build.js -w' 'webpack --context examples --config examples/webpack.config.js -w'",
+    "watch-styleguide": "npm run build && concurrently 'node build.js -w' 'webpack --context examples --config examples/webpack.config.js -w'",
```
tdumitrescu: TBH i don't think anyone uses this; we might as well delete it, now that running `build-examples` is basically preparation for publishing the style guide.

package.json
```diff
@@ -8,7 +8,8 @@
...
+    "watch": "npm run build && concurrently 'node build.js -w'",
```
tdumitrescu: `concurrently` a single thing is just the single thing, no?

antialias: never played with `concurrently` myself; I was just going with what was already there.

package.json
antialias: ok

antialias:
bump @tdumitrescu

package.json
tdumitrescu: just get rid of that superfluous `concurrently` call in the watch task and it'll be good to go

PR #263, opened 2017-05-19
font size is 12px

crmorford:
:|


lib/widgets/tutorial-tooltip/index.styl
```diff
@@ -229,7 +229,7 @@ $tooltip-distance = 12px
     .next, .done
       color blue-100
       cursor pointer
-      font-size 14px
+      font-size 12px
```
zakj: intentional for that to apply to all our existing onboarding tooltips too?

crmorford: hm i added this for notif 3.0 so i assumed we were the only ones using this new feature of it. maybe that was incorrect.

zakj: oh wow I think you're right! how did we not have a "next" before?!

zakj:
0.75.5

PR #264, opened 2017-05-19
PA-1237 tooltip with icon on left

antialias:


antialias:
@mduan fair points. Do you know who in design I should talk to about the warning icon?

mduan:
Eric Lobdell from design should be able to help.

PR #265, opened 2017-05-20
added smart hub components

rickydoar:


lib/widgets/smart-hub-dropdown/index.jade
```diff
@@ -0,0 +1,35 @@
...
+        ) Clear all
+      .mp-sh-menu-divider
+      .mp-sh-alerts-container
+        if alerts.length > 0
+          for alert in alerts
+            mp-smart-hub-alert(
+              attrs={
+                'alert': JSON.stringify(alert),
+              }
+              on={submit: e => $helpers.handleAlertSubmit(e)}
```
mduan: nit: can shorten to `submit: $helpers.handleAlertSubmit`

lib/widgets/smart-hub-dropdown/index.js
```diff
@@ -0,0 +1,106 @@
...
+    this.update({
+      alerts: this.getJSONAttribute(`alerts`),
+      open: this.isAttributeEnabled(`open`),
+    });
+  }
+
+  attributeChangedCallback(name) {
+    super.attributeChangedCallback(...arguments);
+    if (this.initialized) {
+      if (name ===  `open`) {
```
mduan: nit: extra space after `===`

examples/style-guide/index.jade
```diff
@@ -1,5 +1,4 @@
 .style-guide
-
```
mduan: It doesn't look like you've actually added the component to the style guide. If it's easy for you to add it, can you add it and I can play a bit with it. Also, should we be adding to style-guide-new rather than style-guide?

lib/widgets/smart-hub-alert/index.js
```diff
@@ -0,0 +1,110 @@
...
+            units = Math.floor(timestampDiff / 3600000);
+            timeString = units === 1 ? units + ` hour ago` : units + ` hours ago`;
+          } else {
+            units = Math.floor(timestampDiff / 86400000);
+            timeString = units === 1 ? units + ` day ago` : units + ` days ago`;
+          }
+          return timeString;
+        },
+        navigateToReport: () => {
+          this.dispatchEvent(new CustomEvent(`submit`, {
```
mduan: I feel it'd cleaner to split navigateToReport, sendFeedback, and removeAlert as different events, rather than dispatching them all as the `submit` event with different `action`.

lib/widgets/smart-hub-alert/index.js
```diff
@@ -0,0 +1,110 @@
...
+      helpers: {
+        formatAlertTime: timestamp => {
+          const currentTimestamp = new Date().getTime();
+          const timestampDiff = currentTimestamp - timestamp;
+          let timeString;
+          let units;
+          // less than a day ago
+          if (timestampDiff < 86400000) {
+            units = Math.floor(timestampDiff / 3600000);
+            timeString = units === 1 ? units + ` hour ago` : units + ` hours ago`;
```
mduan: Not urgent now, but should switch to using moment.js for these computations at some point.

lib/widgets/smart-hub-dropdown/index.jade
```diff
@@ -0,0 +1,35 @@
...
+    on={
+      change: e => $helpers.handleModalChange('confirmDelete', e.detail.state),
+      submit: e => $helpers.clearAllAlerts(e),
+    }
+  )
+    div
+      .mp-confirm-delete-title Delete all Alerts
+      .mp-confirm-delete-description Are you sure you want to clear all alerts?
+  mp-drop-menu(attrs={open: open}, on={change: e => $helpers.handleDropMenuChange(e)})
+    .mp-sh-menu-container(attrs={'slot-body': true})
```
rickydoar: the mp-drop-menu has a `content` section, so when you put a slot-body attribute on an element within the drop-menu, it knows to slot the element into the content section of the drop-menu

examples/style-guide/index.jade
rickydoar: yea I just put it on style-guide while I was playing around with testing it. I can add it to the new styleguide

lib/widgets/smart-hub-alert/index.js
```diff
@@ -0,0 +1,110 @@
...
+  encodeRison(params) {
+    const risonHash = rison.urlencode(params);
+    return risonHash.substring(1, risonHash.length - 1); // remove leading and trailing parens
+  }
+
+  attachedCallback() {
+    super.attachedCallback();
+    // bootstrap the widget's state from attributes
+    const alert = this.getJSONAttribute(`alert`);
+    this.update({
```
mduan: In terms of the interface of the component. I think you should be trying to model the interface in such a way that it's easiest for the component to work with, rather than molding it to the format of the data returned by the backend. It should be up to the user of the component to get the data into the correct format for the component.

With this in mind, rather than having a single attribute `alert`, I'd make the component have different attributes for each of the pieces of information (message, read, type, etc...) it needs.

rickydoar: This might have been an older comment, but the format should be the same across both

lib/widgets/smart-hub-alert/index.js
rickydoar: In this case I was modeling it after how we handle these types of events in the bookmarks widget. It does feel clean to have it all on submit when actually using the component. Why do you think it would be cleaner to split it out?

lib/widgets/smart-hub-alert/index.js
```diff
@@ -13,100 +12,94 @@ export default registerMPElement(`mp-smart-hub-alert`, class extends Component {
...
+      const units = Math.floor(timestampDiff / 3600000);
+      return units === 1 ? `${units} hour ago` : `${units} hours ago`;
+    } else {
+      const units = Math.floor(timestampDiff / 86400000);
+      return units === 1 ? `${units} day ago` : `${units} days ago`;
+    }
+  }
+
+  generatePermalink() {
+    // TODO(mack): Implement permalink generation
```
rickydoar: This might be hard to do without the reportID available only within mp.report.globals.

mduan: The format returned by the backend will have all the information needed to generate the URL without needing to lookup the bookmark params with the report ID (see https://github.com/mixpanel/analytics/blob/master/clj/service/smart-hub/protos/smart_hub.proto). So it should be possible to generate the permalinks on the front-end.

PR #269, opened 2017-05-23
Event definitions tooltip

jalack:
Adding a friendly tooltip with event definitions so that Lexicon definitions can be shown in Insights, Signal, and anywhere else Panel dropdowns are sold.
screen shot 2017-05-23 at 4 33 27 pm


lib/widgets/items-menu/index.jade
```diff
@@ -1,63 +1,77 @@
...
-            else if item.hasCaret
-              .secondary-action.caret
-                svg-icon(attrs={icon: `caret-right`})
-      if section.isLoading
-        .loading-items
-          mp-spinner
+        if section.isLoading
+          .loading-items
+            mp-spinner
+  mp-event-definition-tooltip(
```
jalack: the diff on this file sucks unfortunately, but the interesting parts to note are this section here, the new top-level element (since there can only be one top-level tag), and some additional handlers for mouseenter and mouseleave events

jbwyme:
we should really consider appending this to the body instead of inline to avoid stacking context issues, I think that's how we need to build all tooltips.

mduan:
I don't feel it makes sense to have the tooltip be part of the `mp-items-menu` widget. The `mp-items-menu` widget is supposed to be generalizable list of items that can be scrollable and navigable, so this feels a little outside the scope of it. Also, the tooltip doesn't actually show inside of the `mp-items-menu` so structurally it feels like a sibling of the `mp-items-menu` rather than a child of it.

Is there anyway you could support having the event definitions tooltip by having the `mp-items-menu` dispatch a `hoverItem` event (possibly with position metadata like the x and y position of the hovered item) and use that for figuring out when to show the tooltip and where to position it?

lib/widgets/event-definition-tooltip/index.js
```diff
@@ -0,0 +1,61 @@
...
+  attachedCallback() {
+    if (this.initialized) {
+      return;
+    }
+
+    if (!this.elementMoved) {
+      this.elementMoved = true;
+      this.parent = this.el.host.parentNode;
+      document.body.style.position = `relative`;
+      document.body.appendChild(this);
```
jalack: your understanding is correct, as best as I can tell. on ted's advice, I started with the way that Insights handles this, which is reflected in the implementation above. but I'm running into the exact issue that you described -- snabbdom inserts a new `mp-event-definition-tooltip` on subsequent updates, resulting in a cascade of tooltips on the end of the body. the way I'm currently handling this is to use snabbdom hooks to kill the old tooltips, but needless to say I don't love that approach. the empty tooltips are innocuous enough, but if we plan on refactoring future tooltips to use this approach would be great to get this ironed out.

the problem with the approach you suggest is that it would render ineffective the `mouseenter` and `mouseleave` events that I need for the tooltip. for those to work the position of the tooltip element must match the content, and that requires it to be on the body. perhaps @tdumitrescu or @jbwyme have more ideas on this one?

jalack: for context here's the Insights PR that uses this component: https://github.com/mixpanel-platform/insights/pull/435

jbwyme: parent snabbdom won't see into the shadowRoot so you can grab stuff from inside your element and append it to the body but taking the host element itself and move it will cause snabbdom to re-add it

jbwyme: i believe you would have two components: the placeholder top level component `<mp-event-definition-tooltip>` and then an internal component (so you can still use panel to build it) `<mp-event-definition-tooltip-internal>` where `<mp-event-definition-tooltip>` just proxys attrs and append `<mp-event-definition-tooltip-internal>` to the body, @tdumitrescu correct me if im wrong.

examples/style-guide-new/components/widgets/index.jade
```diff
@@ -345,5 +345,22 @@
...
                       'sections': JSON.stringify($helpers.getItemsMenuSections()),
-                    },
+                    }
+                    on={
+                      focusItem: ev => ev.detail.item.description ? $helpers.showTooltipForItem(ev.detail.item) : $helpers.hideEventDefinitionTooltip(),
+                      mouseenterItem: ev => $helpers.showTooltipForItem(ev.detail.item),
+                      mouseleaveItem: ev => $helpers.hideEventDefinitionTooltip(),
+                    }
                   )
+                    mp-event-definition-tooltip(
```
jbwyme: this should just be a sibling to `mp-items-menu` rather than a child right?

examples/style-guide-new/components/widgets/index.jade
```diff
@@ -345,5 +345,22 @@
...
+                      on={
+                        mouseenterTooltip: () => $helpers.showEventDefinitionTooltip(),
+                        mouseleaveTooltip: () => $helpers.hideEventDefinitionTooltip(),
+                      }
+                      attrs={
+                        'event-description': eventDefinition ? eventDefinition.description : ``,
+                        'event-name': eventDefinition ? eventDefinition.name : ``,
+                        'lexicon-url': `#`,
+                        show: !!eventDefinition,
```
jbwyme: do we need to support a `show` attr or can we just do:
```
if eventDefinition
  mp-event-definition-tooltip
      ...
```
?

examples/style-guide-new/components/widgets/index.jade
```diff
@@ -345,5 +345,22 @@
                     attrs={
                       'search-filter': itemsMenuSearchFilter,
                       'sections': JSON.stringify($helpers.getItemsMenuSections()),
-                    },
+                    }
+                    on={
+                      focusItem: ev => ev.detail.item.description ? $helpers.showTooltipForItem(ev.detail.item) : $helpers.hideEventDefinitionTooltip(),
```
jbwyme: would it be a simpler api to just have `mp-items-menu` dispatch `blurItem` as well and then you can get rid of the `mouseenterItem` and `mouseleaveItem` events completely?

Related: when the menu closes, we need to close the tooltip too. If we take the `blurItem` approach we can trigger that as well on menu close

http://cometapp.us/c/hnlpNvvU

examples/style-guide-new/components/widgets/index.jade
```diff
@@ -345,5 +345,22 @@
...
+                      mouseenterItem: ev => $helpers.showTooltipForItem(ev.detail.item),
+                      mouseleaveItem: ev => $helpers.hideEventDefinitionTooltip(),
+                    }
                   )
+                    mp-event-definition-tooltip(
+                      on={
+                        mouseenterTooltip: () => $helpers.showEventDefinitionTooltip(),
+                        mouseleaveTooltip: () => $helpers.hideEventDefinitionTooltip(),
```
jbwyme: is this actually the behavior we want? it results in a UX of this: http://cometapp.us/c/JESyHQKD

lib/widgets/event-definition-tooltip/index.js
```diff
@@ -0,0 +1,56 @@
...
+        posTop: `0`,
+        show: false,
+      },
+      useShadowDom: true,
+    };
+  }
+
+  attachedCallback() {
+    super.attachedCallback(...arguments);
+    this.parent = this.el.host.parentNode;
```
jbwyme: just curious, how does `this.el.host.parentNode` differ from `this.parentNode`? I.e. what is `this.el.host`?

lib/widgets/event-definition-tooltip/index.js
```diff
@@ -0,0 +1,56 @@
...
+
+  detachedCallback() {
+    document.body.removeChild(this.domPatcher.vnode.elm);
+    this.domPatcher = null;
+    super.detachedCallback(...arguments);
+  }
+
+  attributeChangedCallback(name, oldValue, newValue) {
+    super.attributeChangedCallback(...arguments);
+    if (name === `event-name`) {
```
jbwyme: would prefer a switch here since it's all checking `name`

lib/widgets/event-definition-tooltip/index.js
```diff
@@ -0,0 +1,56 @@
...
+    };
+  }
+
+  attachedCallback() {
+    super.attachedCallback(...arguments);
+    this.parent = this.el.host.parentNode;
+  }
+
+  detachedCallback() {
+    document.body.removeChild(this.domPatcher.vnode.elm);
```
jbwyme: this seems sort of magical... would it be possible to set the `elm` as a part of a snabbdom hook in the jade file?

lib/widgets/event-definition-tooltip/index.js
```diff
@@ -0,0 +1,56 @@
...
+
+import {registerMPElement} from  '../../util/register-element.js';
+import './internal';
+
+import template from './index.jade';
+
+import css from './index.styl';
+
+// We don't know the exact height of the tooltip at the time we position it, so this serves as an approximate placeholder
+const TOOLTIP_HEIGHT_IN_PX = 152;
```
jbwyme: why can't we use the snabbdom hook `insert` to delay measuring until then?

lib/widgets/event-definition-tooltip/internal/index.js
```diff
@@ -0,0 +1,50 @@
...
+        show: false,
+      },
+      useShadowDom: true,
+    };
+  }
+
+  attachedCallback() {
+    if (!this.elementMoved) {
+      this.elementMoved = true;
+      document.body.style.position = `relative`;
```
jbwyme: i don't think we want to manipulate the body styles, the body should already effectively be relatively positioned anyways.

lib/widgets/event-definition-tooltip/index.jade
```diff
@@ -0,0 +1,14 @@
+- const url = $component.hasAttribute('lexicon-url') ? $component.getAttribute('lexicon-url') : `#`
+mp-event-definition-tooltip-internal(
```
jbwyme: my original thought was that this isn't in the jade at all. Instead this file would be basically an empty div. then in the `index.js` you'd just append to the body in `attachedCallback` like 
```js

attachedCallback() { 
  this.tooltip = document.createElement('mp-event-definition-tooltip-internal');
  this.tooltip.setAttribute('blah', 'val');
  document.body.appendChild(this.tooltip);
}

detatchedCallback() {
  document.body.removeChild(this.tooltip);
}
```

To me, that makes the logic a little easier/clear than what we currently have for appending to the body. What are your thoughts?

examples/style-guide-new/components/widgets/index.jade
jalack: followed the existing pattern for tooltips where they tend to be children of the item they're pointing to. parent node has easy access and given that the element is on the body it's not a big difference to me either way

examples/style-guide-new/components/widgets/index.jade
jalack: basically, yes. without some kind of ability to keep the tooltip open on hover, it's pretty much impossible to click the Lexicon link. open to other suggestions on how we could achieve that though

examples/style-guide-new/components/widgets/index.jade
jbwyme: yeah but i don't think `mp-items-menu` supports children which makes it odd

lib/widgets/event-definition-tooltip/index.js
jalack: looks like they refer to the same thing, ie `this.el.host === this`. not sure why I had this particular syntax

examples/style-guide-new/components/widgets/index.jade
jalack: fair point, I'll change

examples/style-guide-new/components/widgets/index.jade
jalack: I don't think we need to support a `show` attribute, but we'll want to leave the tooltip in the DOM to facilitate animations

lib/widgets/event-definition-tooltip/index.jade
jalack: I don't feel terribly strongly -- on the one hand, I like how my current approach makes updating the styles/attributes and attaching event handlers pretty intuitive, and aligns with how we do things elsewhere. but there's definitely a benefit to encapsulating all the DOM manipulation messiness in one file to make it more understandable. I could go either way, though I tend to think that what's currently here might make for an easier reuse pattern elsewhere, if only because you just have to copy a small bit of magic and then go about building as you would otherwise

lib/widgets/event-definition-tooltip/index.js
jalack: the snabbdom hooks don't seem to be firing correctly as far as I can tell in my testing -- can make this a bit less magical though

lib/widgets/event-definition-tooltip/index.js
jalack: the `insert` hook isn't firing on `mp-tooltip-internal` -- still investigating why but the height isn't available on `postpatch` either. one advantage of this hack is that the tooltip doesn't jump around as its height changes, which is something that Suhail has asked from us. I'll add a note in the comment that such behavior is desired

examples/style-guide-new/components/widgets/index.jade
jalack: good catch on the menu close bit. I may be misunderstanding, but my concern would be that I wanted to keep the tooltip & items menu decoupled as much as possible. I don't see how we could get rid of `mouseenterItem` and `mouseleaveItem` events without introducing some event (`focusItem` or otherwise) that's specific to what the tooltip needs.

examples/style-guide-new/components/widgets/index.jade
```diff
@@ -345,5 +345,20 @@
...
+                    on={
+                      focusItem: ev => ev.detail.item.description ? $helpers.showTooltipForItem(ev.detail.item) : $helpers.hideEventDefinitionTooltip(),
+                      mouseenterItem: ev => $helpers.showTooltipForItem(ev.detail.item),
+                      mouseleaveItem: ev => $helpers.hideEventDefinitionTooltip(),
+                    }
+                  )
+                  mp-event-definition-tooltip(
+                    on={
+                      mouseenterTooltip: () => $helpers.showEventDefinitionTooltip(),
+                      mouseleaveTooltip: () => $helpers.hideEventDefinitionTooltip(),
```
jbwyme: > jbwyme 19 hours ago Owner
> is this actually the behavior we want? it results in a UX of this: http://cometapp.us/c/JESyHQKD
> 
> jalack 18 hours ago Member
> basically, yes. without some kind of ability to keep the tooltip open on hover, it's pretty much impossible to click the Lexicon link. open to other suggestions on how we could achieve that though

You just don't remove the tooltip on mouseleave. I.e. as long as the element in the menu is active just continue showing the tooltip

lib/widgets/event-definition-tooltip/index.js
```diff
@@ -0,0 +1,60 @@
...
+import template from './index.jade';
+
+import css from './index.styl';
+
+/**
+ * We don't know the exact height of the tooltip at the time we position it, so this serves as an approximate placeholder.
+ * Although this leads to imperfect vertical centering, it prevents the tooltip's vertical position from changing based
+ * on its height, which is desired behavior.
+ */
+const TOOLTIP_HEIGHT_IN_PX = 152;
```
jbwyme: > jbwyme 19 hours ago Owner
> why can't we use the snabbdom hook insert to delay measuring until then?
> 
> jalack an hour ago Member
> the insert hook isn't firing on mp-tooltip-internal -- still investigating why but the height isn't available on postpatch either. one advantage of this hack is that the tooltip doesn't jump around as its height changes, which is something that Suhail has asked from us. I'll add a note in the comment that such behavior is desired

This might be due to how you're manipulating the component. I suspect if you do the `document.body.appendChild` approach it'll fire.

examples/style-guide-new/components/widgets/index.jade
jalack: I like that idea -- would make everything easier. I'll follow up with Tim on this one

lib/widgets/event-definition-tooltip/index.jade
jalack: no need, after implementing it I think I'm inclined to agree with you

lib/widgets/event-definition-tooltip/internal/index.jade
```diff
@@ -0,0 +1,12 @@
+- const url = $component.getAttribute('url')
+.mp-event-definition-wrapper
+  if show
```
jbwyme: i'm still a little unclear on why we don't just let the caller determine the tooltip being visible based on a conditional in the jade. Then make animations work with add/remove from dom?

examples/style-guide-new/components/widgets/index.jade
```diff
@@ -345,5 +345,15 @@
                     attrs={
                       'search-filter': itemsMenuSearchFilter,
                       'sections': JSON.stringify($helpers.getItemsMenuSections()),
-                    },
+                    }
+                    on={
+                      focusItem: ev => ev.detail.item.description ? $helpers.showTooltipForItem(ev.detail.item) : $helpers.hideEventDefinitionTooltip(),
+                      mouseenterItem: ev => ev.detail.item.description ? $helpers.showTooltipForItem(ev.detail.item) : $helpers.hideEventDefinitionTooltip(),
```
jbwyme: this can be removed, focusItem should already be called when the item receives mouseenter

jalack: the code seems to imply that, but `this.focusItemIfNeeded()` isn't invoked even though the focus index is changed. idk if focus is meant to be a distinct state from hover, but we could invoke the `focusItem` event here if they are really the same concept in this case

jbwyme: @mduan ?

lib/widgets/event-definition-tooltip/internal/index.jade
jalack: to allow the exit animation to complete, we need to be able to delay the removal of the tooltip element. we could do the same here, but then each caller would be responsible for providing its own animation out for the tooltip. I'd rather the component maintain responsibility for exit animation and handle removing anything visible from the DOM

examples/style-guide-new/components/widgets/index.jade
mduan: `focusItemIfNeeded()` is probably not the best name for that method. What it's doing is moving the focused item into the viewport in the case of a scrollable list. This is needed when you use keyboard navigation (up, down) to focus an item, but not when you use the mouse hover to focus an item. A better name for it might be `scrollMenuIfNeeded`.

I agree with Josh that the most useful events for the `mp-items-menu` to fire would be `focusItem` and `blurItem`. There isn't really a difference between `focusItem` and `mouseenterItem` from the caller's perspective.

lib/widgets/items-menu/index.js
```diff
@@ -250,6 +251,8 @@ class ItemsMenu extends Component {
     } else if (focusedElBottom + THRESHOLD_BOTTOM > listElBottom) {
       listEl.scrollTop = focusedElBottom + THRESHOLD_BOTTOM - listEl.offsetHeight;
     }
+
```
jbwyme: let's go ahead and rename this method to `scrollToItemIfNeeded()`

lib/widgets/items-menu/index.js
```diff
@@ -62,6 +62,7 @@ class ItemsMenu extends Component {
             isPropertiesPillFocused: false,
             focusedItemIndex: item.index,
           });
+          this.dispatchEvent(new CustomEvent(`focusItem`, {detail: {item}}));
```
jbwyme: i think the slightly more correct thing here would be to create a method called `focusItem(itemIdx)` that updates the state with the focusedItemIndex and also dispatches this event. That way you can also get it out of `focusItemIfNeeded` and rename it to `scrollToItemIfNeeded()` 

Then you can call your new method here: https://github.com/mixpanel/mixpanel-common/blob/fc3c0d3de7097e879ec8100698b319a7defcf914/lib/widgets/items-menu/index.js#L105

lib/widgets/event-definition-tooltip/index.js
```diff
@@ -0,0 +1,51 @@
...
+    super.attachedCallback(...arguments);
+    this.tooltip = document.createElement(`mp-event-definition-tooltip-internal`);
+    this.tooltip.setAttribute(`url`, this.getAttribute(`lexicon-url`) || `#`);
+    document.body.appendChild(this.tooltip);
+    this.positionElement = this.previousSibling;
+  }
+
+  detachedCallback() {
+    document.body.removeChild(this.tooltip);
+    this.domPatcher = null;
```
jbwyme: is this needed?

lib/widgets/event-definition-tooltip/index.styl
```diff
@@ -0,0 +1,5 @@
+@import '../../stylesheets/defaults'
```
jbwyme: can't we just remove this file completely now?

lib/widgets/event-definition-tooltip/internal/index.js
```diff
@@ -0,0 +1,50 @@
...
+import css from './index.styl';
+
+export default registerMPElement(`mp-event-definition-tooltip-internal`, class extends Component {
+  get config() {
+    return {
+      css,
+      template,
+      defaultState: {
+        delayRemove: timing[`mp-fast`],
+        show: false,
```
jbwyme: i know that i requested you remove the `show` attribute as a part of my request to just allow the client handle hiding/showing but since we cannot let the client do that (because of animations) let's add back the explicit attribute instead of implicitly doing it based on `eventDefinition`. Let's also call it `open` to keep consistency with `mp-drop-menu`

PR #270, opened 2017-05-25
Updates to the mp-items-menu and added mp-icon-input

jbwyme:


lib/widgets/items-menu/index.js
```diff
@@ -74,17 +74,16 @@ class ItemsMenu extends Component {
...
-    super.attachedCallback(...arguments);
-
-    this.handleScroll = ev => {
-      const scrollBottom = ev.target.scrollTop + ev.target.offsetHeight;
-      if (this.state.isProgressive && scrollBottom + PROGRESSIVE_LIST_BUFFER_PX >= ev.target.scrollHeight) {
-        this.increaseProgressiveListSize();
-      }
-    };
-    this.addEventListener(`scroll`, this.handleScroll);
+  set keydownProxy(domElement) {
```
jbwyme: can be used in host jade like:
```
mp-items-menu(props={keydownProxy: $component})
```

mduan: Can't think of a better name for it right now, but what about appending "El"? So `keydownProxyEl`.

tdumitrescu: re: naming, I'd try to make it something that makes it obvious that it's an element, even as simply as `keydownProxyEl`. My suggestions: `keyboardEl`, `keyboardListenerEl`

lib/widgets/items-menu/index.js
```diff
@@ -108,28 +115,46 @@ class ItemsMenu extends Component {
...
+      case `search-filter`:
+        this.update({searchFilter: newVal});
+        break;
+      case `empty-message`:
+        this.update({emptyMessage: newVal});
+        break;
+      case `sections`:
+        this.update({sections: this.getJSONAttribute(`sections`)});
+        break;
+      case `progressive`:
```
mduan: This change is not backwards compatible with existing usages of `mp-items-menu`, so remember to add the progressive attr to callers when you publish this PR.

lib/widgets/items-menu/index.js
```diff
@@ -226,13 +251,17 @@ class ItemsMenu extends Component {
     }
   }
 
-  handleClickItem(item) {
-    this.dispatchEvent(new CustomEvent(`clickItem`, {detail: {item: this.unwrapItem(item)}}));
+  handleClickItem(item, ctrlKey = false) {
+    this.dispatchEvent(new CustomEvent(`clickItem`, {detail: {item: this.unwrapItem(item), ctrlKey}}));
```
mduan: rename `ctrlKey` to `modifierKey`?

PR #272, opened 2017-05-26
tooltip tweak so that you can click right next to a tooltip

ldelbeccaro:
the arrow created from the ::after element blocks clicking on something it may be pointing to


ldelbeccaro:
<3

PR #273, opened 2017-05-30
Adding mp-toggle-option disabled attribute

nojvek:
image


lib/components/toggle/index.js
```diff
@@ -19,6 +19,7 @@ export default registerMPElement(`mp-toggle`, class extends Component {
           .map(el => ({
             text: el.textContent,
             value: el.getAttribute(`value`),
+            disabled: !!el.getAttribute(`disabled`),
```
tdumitrescu: Might be worth considering whether it's worth using `el.isAttributeEnabled('disabled')`, which allows for the usage `<mp-toggle-option disabled="false">`. We tend to do this for custom attributes in our components, to make it easier to manage from vdom libraries. But...`disabled` as a standard html attr is a bit of a special case - on `button`s and other elements its mere presence causes disabling, regardless of its value. So it may be best for this one just to check for the presence of `disabled` like this. I could go either way.

nojvek: Didn't realize that was a thing. Thank you 😄  using `isAttributeEnabled`

nojvek:
Don't merge this yet. Just found another bug.

nojvek:
Okay ready to be merged and published.

PR #276, opened 2017-05-31
tooltip hides as expected when user does mouseout before next animation frame

antialias:


antialias:
doesn't look like the failing specs are related to my change:

```Windows 10 microsoftedge latest ✖ test/ » relativeToAbsoluteDate » converts a relative date integer and a unit to the expected Date object
  expected 1759190400000 to sort of equal 1759276800000
    <unknown> at   Assertion.prototype.assert at build/suites.js:486:0
    <unknown> at      Assertion.prototype.eql at build/suites.js:614:0
    <unknown> at           Anonymous function at build/suites.js:28630:0
    <unknown> at           Anonymous function at build/suites.js:28621:0
Windows 10 microsoftedge latest ✖ test/ » normalizeDateStrings » allows utc offset to define the current moment
  expected [ '2016-06-01', '2017-05-31' ] to sort of equal [ '2017-05-31', '2017-06-01' ]
    <unknown> at   Assertion.prototype.assert at build/suites.js:486:0
    <unknown> at      Assertion.prototype.eql at build/suites.js:614:0
    <unknown> at           Anonymous function at build/suites.js:28691:0
Windows 10 microsoftedge latest Tests failed: 2 failed tests
```

antialias:
if someone could help me figure out how to run sauce tests from my devbox, I could find where these started failing.

evnp:
@antialias not sure about sauce, but I'm having luck with `npm run unit-test` (and only the 2 tests enabled). Anyway I'll have a fix up shortly

test/components/tooltip.js
```diff
@@ -0,0 +1,39 @@
+import bubble from 'bubble-event';
+import createMockRaf from 'mock-raf';
+import '../../lib/components/tooltip';
+
+describe('tooltip', function () {
+  let t;
+  let clock;
```
mduan: Is clock needed, I don't see it being used anywhere.

test/components/tooltip.js
```diff
@@ -0,0 +1,39 @@
...
+  let mockRaf;
+  let sandbox;
+  beforeEach(function (done) {
+    sandbox = sinon.sandbox.create();
+    mockRaf = createMockRaf();
+    clock = sinon.useFakeTimers();
+    t = document.createElement(`mp-tooltip`);
+    document.body.appendChild(t);
+    mockRaf.step();
```
mduan: Is this necessary? `requestAnimationFrame` isn't stubbed until line 19.

test/components/tooltip.js
```diff
@@ -0,0 +1,39 @@
...
+    t.update()
```
mduan: Is this update necessary? There's no state being updated, so it doesn't seem like this update will do anything. Also, missing semi-colon.

test/components/tooltip.js
```diff
@@ -0,0 +1,39 @@
+import bubble from 'bubble-event';
```
mduan: nit: It seems simple enough to call `dispatchEvent`. Is there a reason to have a import a library for this?

antialias:
> I'm having luck with `npm run unit-test`

Looks like I'm able to reproduce the failing specs locally too, so maybe they aren't edge-specific.

test/components/tooltip.js
```diff
@@ -0,0 +1,39 @@
...
+
+describe('tooltip', function () {
+  let t;
+  let clock;
+  let mockRaf;
+  let sandbox;
+  beforeEach(function (done) {
+    sandbox = sinon.sandbox.create();
+    mockRaf = createMockRaf();
+    clock = sinon.useFakeTimers();
```
antialias: wct

test/components/tooltip.js
antialias: yes, this spec addresses an issue with a raf callback in the tooltip code.

test/components/tooltip.js
antialias: It keeps the specs concise. The current DOM event API requires a custom event to be created and dispatched to the node. The alternative would look something like this:

```js
toolipEl.dispatchEvent(new CustomEvent('mousenter'));
toolipEl.dispatchEvent(new CustomEvent('mouseleave'));
```

The problem is that [`CustomEvent`](https://developer.mozilla.org/en-US/docs/Web/API/CustomEvent/CustomEvent#Browser_compatibility) is not available in IE (it has `document.createEvent` instead), so [`bubble-event`](https://github.com/1stdibs/bubble-event/blob/master/index.js) includes the [`custom-event-universal`](https://www.npmjs.com/package/custom-event-universal) ponyfill that takes care of this.

`bubble-event` is specified as a dev dep, so it won't contribute to the production bundle size.

tdumitrescu: FWIW My preference is to keep these in-browser tests more realistic in terms of using the actual browser APIs that the non-test code uses, so real `requestAnimationFrame`, real `dispatchEvent` etc. We use `new CustomEvent` all over the place in the component code, so if it doesn't work there then it shouldn't work in tests either.

antialias: If `-common` does not support IE at all, which appears to be the case, then `new CustomEvent` will work just fine.

antialias: > My preference is to keep these in-browser tests more realistic in terms of using the actual browser APIs

@tdumitrescu : The two specs that I'm adding here do not intend to test browser integration. Perhaps there is a better place for unit-style specs? I would prefer to keep mocking `raf` in order to keep the spec simple. Specifically, I'd like to avoid writing async code here.

lib/components/tooltip/index.js
```diff
@@ -39,7 +39,12 @@ export default registerMPElement(`mp-tooltip`, class extends Component {
       placement,
     });
 
+    const updateRef = this.lastUpdateRef = {};
     window.requestAnimationFrame(() => {
+      if (this.lastUpdateRef !== updateRef) {
+        // another call to show or hide has occurred since thse invocation that created this closure
```
nojvek: `thse` spelling error

lib/components/tooltip/index.js
```diff
@@ -39,7 +39,12 @@ export default registerMPElement(`mp-tooltip`, class extends Component {
       placement,
     });
 
+    const updateRef = this.lastUpdateRef = {};
```
nojvek: What does updateRef supposed to hold? Does it need to be an object if you're just doing an equality check. Would prefer comments explaining the logic.

test/components/tooltip.js
tdumitrescu: IMO if you're relying on event-firing and requestAnimationFrame, you're writing a browser test and ought to run it in a real browser environment, which we have infrastructure in place for. If you want to be able to unit test parts of a component without the integrated environment, I'd suggest extracting the relevant portions into their own methods/functions which can be called without the global browser context expected in the current code; that way you can run it server-side without mocking a browser. There is also limited support for rendering Panel components server-side into HTML strings, but I don't think that's useful for these interaction tests.

lib/components/tooltip/index.js
antialias: `updateRef` doesn't hold data. A few lines down `this.lastUpdateRef` is compared to `updateRef`, and if it's different, that means another update has occurred since the call in which `updateRef` was created.

lib/components/tooltip/index.js
```diff
@@ -84,6 +89,7 @@ export default registerMPElement(`mp-tooltip`, class extends Component {
   }
 
   hide() {
+    this.lastUpdateRef = {};
```
nojvek: isn't a true false boolean going to work here? I still don't understand why you need a new instance of an object.

test/components/tooltip.js
antialias: I agree that there is virtue in a spec running in as authentic an environment as possible, but for what this code is actually testing, it's not necessary to rely on the browser's implementation of `raf` in order to test the actual fix. IMO, how specs are written should be evaluated on a case-by-case basis.

I've gone ahead and removed the `raf` mock though, so I hope that' sufficient to get a 👍  on this PR.

lib/components/tooltip/index.js
antialias: Discussed in person and learned about `window.cancelAnimationFrame`, so using that instead.

lib/components/tooltip/index.js
```diff
@@ -31,65 +31,68 @@ export default registerMPElement(`mp-tooltip`, class extends Component {
...
-      let parentOffsetTop;
-      const parentPosition = window.getComputedStyle(this.parentNode).getPropertyValue(`position`);
-      // The tooltip has position absolute, which means it will be positioned relative to the nearest ancestor with
-      // non-static position. Different handling is needed depending on whether the direct parent of the tooltip
-      // has non-static positioning.
-      if (parentPosition === `static`) {
-        parentOffsetLeft = this.parentNode.offsetLeft;
-        parentOffsetTop = this.parentNode.offsetTop;
+    window.cancelAnimationFrame(this.lastRaf);
+    this.lastRaf = window.requestAnimationFrame(this._showAfterRAF.bind(this));
```
nojvek: how about `_rafId` ?

lib/components/tooltip/index.js
```diff
@@ -31,65 +31,68 @@ export default registerMPElement(`mp-tooltip`, class extends Component {
...
+  }
+  _showAfterRAF() {
```
nojvek: `_computePosition` ?

antialias: Looks like the code does more than compute the position; I'd prefer to keep the name I gave it.

lib/components/tooltip/index.js
antialias: Good point about using an underscore, but I think `_lastRaf` is more descriptive.

test/components/tooltip.js
```diff
@@ -0,0 +1,42 @@
...
+  let tooltipEl;
+  beforeEach(function (done) {
+    tooltipEl = document.createElement(`mp-tooltip`);
+    document.body.appendChild(tooltipEl);
+    window.requestAnimationFrame(() => {
+      done();
+    });
+  });
+  it('should only call the raf callback once', function (done) {
+    const sandbox = sinon.sandbox.create();
```
mduan: Do you think you could add sinon to package.json and import it here? That way, we can have more explicit control over which version is used, and it removes confusion from other people looking at the code and wondering where sinon is coming from.

tdumitrescu: not sure i agree with trying to override the tooling provided by the testing framework, until/unless we have a real need for it

mduan: Hm , fair enough. I suppose we are `beforeEach`, `describe`, and `it` provided by the testing framework which are globals. It's just than `sinon` is used rarely enough right now that it's not obvious where it's coming from.

I also realized we don't have linting for the `test/` directory. Otherwise, there'd need to be `/* global ... */` comment to use these globals.

antialias: > It's just than `sinon` is used rarely enough right now

What do you guys do for stubbing / mocking and spying?

PR #280, opened 2017-05-31
Date util bugfix

evnp:
Logic in `parseDate` is bad - the string "2017 06 01T10 39 22" parses to `Wed Jun 01 2016 00:00:00 GMT-0700 (PDT)`. This is because of some messy logic that attempts to check whether year was specified in the input string, necessary because
```
> moment('12/31', 'M D')
moment("2017-12-31T00:00:00.000")
```
parses to the future when we want it to parse to past date. The check can be simplified by checking if `moment('12/31', 'M D', true)` (true for "strict parsing" mode) is valid, much better than messy string manipulation that can never be fully correct.

Not exactly sure why the `relativeToAbsoluteDate` test broke, seems like a TZ issue. Setting up the reference data with `moment` instead of `Date` (as it should be anyway, think the test was written before we used moment in the tests) fixes the issue.


lib/util/date.js
```diff
@@ -81,14 +81,7 @@ export function parseDate(dateString, {iso=false, startOfDay=false, endOfDay=fal
...
       // make dates like "12/31" parse to the most recent past instance of that date
-      const fullYear = mDate.format(`YYYY`);
-      const partialYear = mDate.format(`YY`);
-      const month = mDate.format(`MM`);
-      const day = mDate.format(`DD`);
-      if ( // was the year specified?
-        !dateString.includes(fullYear) ||
-        !dateString.replace(month, ``).replace(day, ``).includes(partialYear) // ensure we don't mistakenly match on month or day
-      ) {
+      if (moment(dateString, `M D`, true).isValid()) {
```
jbwyme: are you sure this is simpler? 😄

evnp: fair... I'll add a comment explaining it

jbwyme: oh i was kidding but thanks!

PR #284, opened 2017-05-31
build a js version of json files so they can be imported into js file…

jbwyme:
…s without a loader


build.js
```diff
@@ -95,16 +104,21 @@ function transpileFile(filename) {
...
-  let result = transpiler(filename);
-  if (result.filename) {
-    filename = result.filename;
-    result = result.output;
-  }
-  const outputFile = filename.replace(INPUT_DIR, OUTPUT_DIR);
-  ensureDir(path.dirname(outputFile));
+  const transpilers = typeof transpiler === 'function' ? [transpiler] : transpiler;
+
+  transpilers.forEach(transpiler => {
```
tdumitrescu: this new param shadows `transpiler` in the outer context. how about renaming the outer one to `transpilers` and doing this:
```js
if (typeof transpilers === `function`) {
  transpilers = [transpilers];
}
```

build.js
```diff
@@ -58,7 +58,16 @@ const CONFIGS = [
   {
     matches: /lib\/stylesheets\/mixins/,
     transpilers: {
-      json: passThrough,
+      json: [
+        passThrough, // straight json for importing into stylus
+        filename => { // js for importing into js files with no additional loaders
```
tdumitrescu: This block specifically transforms only `lib/stylesheets/mixins`; do we want to do anything with json files in other source dirs?

build.js
```diff
@@ -58,7 +58,16 @@ const CONFIGS = [
...
+          const js = fs.readFileSync(filename).toString().replace(/\n/g, ' ');
```
tdumitrescu: why do we need to replace newlines in the JSON? this seems like legit output to me:
```js
modules.exports = {
  "mp-slow": 400,
  "mp-fast": 250
};
```
and will help debugging in dev mode. production bundles have a separate minification flow.

.babelrc
```diff
@@ -1,7 +1,6 @@
 {
   "plugins": [
     "babel-plugin-import-glob",
-    ["inline-json-import", {}],
```
tdumitrescu: w00t

PR #288, opened 2017-06-05
Add user icon widget

jalack:
Currently used in Notifications 3.0 with a TODO, needs to be added to Lexicon as well so converting to a common component for all to enjoy.

screen shot 2017-06-05 at 1 29 51 pm


lib/widgets/user-icon/index.jade
```diff
@@ -0,0 +1,5 @@
+.mp-user-icon-container
+  if name
+    div(class={[`mp-user-icon-color-${name[0].toUpperCase()}`]: true})
+      = $helpers.nameToInitials(name)
+      mp-tooltip= $helpers.getAbbreviatedName()
```
jbwyme: any chance you'd make this configurable? in my case I don't want a tooltip

jbwyme: im also guessing some implementations will need a different tooltip text so maybe make this the default, allow to pass in a different tooltip text or turn it off completely

tdumitrescu:
Leaving no good deed unpunished: can you make it work in Insights and JQL Console, both of which also have this tooltip?

lib/widgets/user-icon/index.js
```diff
@@ -0,0 +1,33 @@
...
+      template,
+      useShadowDom: true,
+      defaultState: {
+        name: null,
+      },
+      helpers: {
+        nameToInitials,
+        getAbbreviatedName: () => {
+          const nameParts = this.state.name.split(` `);
+          return `${nameParts[0]} ${nameParts[nameParts.length - 1][0]}.`;
```
tdumitrescu: Is a full multi-part name always available?
screen shot 2017-06-05 at 2 17 28 pm
(from jql console)

PR #292, opened 2017-06-06
New chrome updates

jbwyme:


lib/stylesheets/mixins/icons.styl
```diff
@@ -1,23 +1,45 @@
...
   svg-icon
-    if size != null
+    display inline-block
+    height size
+    min-height size
+    min-width size
+    position relative
+    width size
+
+    &[icon=type-boolean]
```
crmorford: moving these here means we have to rely on using svg-icon-style() for every boolean/lightning-bolt/alert icon even if the default color/size is fine (which feels weird to me as a general rule, although it seems like it would almost always work the way we use icons).

lightning bolt should be fine just because it always gets changed to purple -- i'd wanna do a lot of QA to `alert` and `boolean` icons across analytics/dashboards/insights before upgrading to this.

crmorford: oh i see -- you're calling it in `lib/stylesheets/defaults/svg-icon.styl` 
..never mind

PR #294, opened 2017-06-06
input shift safari

crmorford:
icon-inputs work on safari (and ff)

not foolproof. this just makes it able to override all the other styles attached to `input` (which we have a suprising amount of in mixpanel common).


lib/stylesheets/mixins/icons.styl
```diff
@@ -30,7 +30,7 @@ icon-input(icon-color=grey-500, icon-color-focus=grey-700)
   display inline-flex
   position relative
 
-  input
+  input.shifted-input
```
zakj: maybe a comment above this line that the class is needed until we have proper css containment; otherwise someone else is possibly going to come along and see this "unnecessary" class, remove it, test it in Chrome, and pat themselves on the back

crmorford: ok can do!

lib/stylesheets/mixins/icons.styl
```diff
@@ -30,7 +30,8 @@ icon-input(icon-color=grey-500, icon-color-focus=grey-700)
   display inline-flex
   position relative
 
-  input
+  // this class helps encapsulate styles, in browsers that don't support shadow dom
+  input.shifted-input
```
jbwyme: i believe there are some places uses this mixin that will break with this additional selector

jbwyme: also we may as well follow our verbose pattern since we're trying to encapsulate: `mp-icon-input-input` :(

crmorford: so as a general rule, i should use `.<component-name>-<element>`?
just wanna make sure the double 'input' in the class name you proposed was intentional.

jbwyme: yeah what I've been doing is using the `prefix-classes` mixin like this: https://github.com/mixpanel/mixpanel-common/blob/master/lib/components/button/index.styl#L31

that way you don't have to type the prefix every time in the `styl` files but unfortunately still have to in the jade files

jbwyme: actually i have a better idea to solve this problem: don't modify the mixin, copy these styles into the `mp-icon-input` stylus file and modify away. then it's backwards compatible and we can deprecate the mixin.

PR #298, opened 2017-06-07
Tooltip spic & span polish

jalack:
- Removing translation on animation
- Shadow tweaks
- Use UTC time for verification dates
- Better examples/discovery in the style guide
- Ability to copy/paste content within tooltip
- Readability improvements
- Copy tweaks
- Green icons for verified events
- Close tooltip & drop menu simultaneously

screen shot 2017-06-08 at 1 43 02 pm


examples/style-guide-new/components/widgets/index.js
```diff
@@ -178,9 +182,12 @@ document.registerElement(`widgets-section`, class extends Component {
               isDisabled: true,
               hasPropertiesPill: true,
             }, {
-              label: `[Collect everything] Clicked cancel edit elements`,
+              label: `Clicked cancel edit elements`,
```
tdumitrescu: How come you're shortening? Hope it's not because it looks bad...

examples/style-guide-new/index.js
```diff
@@ -45,10 +45,6 @@ document.registerElement(`style-guide`, class extends Component {
         searchTermPrompt: null,
         searchTerm: null,
         searchSections: [],
-        sectionFilter: null,
-        sectionOpen: `colors`,
-        subSectionOpen: null,
-        SVG_ICONS,
```
tdumitrescu: this is why people resort to alphabets

examples/style-guide-new/components/widgets/index.jade
```diff
@@ -339,7 +339,10 @@
               .drop-menu-container
                 mp-drop-menu(
                   attrs={open: itemsMenuOpen}
-                  on={change: $helpers.handleItemsMenuModalChange}
+                  on={
+                    change: $helpers.handleItemsMenuModalChange,
+                    closing: $helpers.handleItemsMenuModalClosing,
```
tdumitrescu: why do these handlers refer to "ItemsMenuModal"? isn't this a dropdown?

lib/widgets/event-definition-tooltip/internal/index.jade
```diff
@@ -1,10 +1,12 @@
...
-    .mp-event-definition-tooltip(class={
-      delayed: {'mp-event-definition-in': true},
-      remove: {'mp-event-definition-out': true, delayRemove},
-    })
+    .mp-event-definition-tooltip(
+      class={
+        delayed: {'mp-event-definition-in': true},
+        remove: {'mp-event-definition-out': true, delayRemove},
+      }
+    )
```
tdumitrescu: why

lib/widgets/event-definition-tooltip/internal/index.js
```diff
@@ -33,11 +34,20 @@ export default registerMPElement(`mp-event-definition-tooltip-internal`, class e
       document.body.appendChild(this);
     }
 
+    // prevent menu close when the user clicks inside the tooltip
+    this.clickInsideListener = e => {
+      if (!clickWasOutside(e, this)) {
+        e.stopImmediatePropagation();
```
tdumitrescu: so whenever one of these tooltips is in the DOM, clicking on it will always prevent click-outside for any other DOM element?

examples/style-guide-new/index.js
jalack: I believe that this is why people use linters...

lib/widgets/event-definition-tooltip/internal/index.jade
jalack: readability

jalack: readability

lib/widgets/event-definition-tooltip/internal/index.js
jalack: yes, without tightly coupling it to the drop-menu element I'm not sure how to obtain the desired result. happy to hear a suggestion though since I am wary of stopping propagation of events in a global way

examples/style-guide-new/components/widgets/index.js
jalack: nah, looks fine -- I found it a bit confusing that it was prefixed with [Collect Everything] but wasn't an autotrack event. I understand that it's a regular event for the collect everything product, but I think it's clearer that it's an event without the prefix

lib/components/drop-menu/index.js
```diff
@@ -35,6 +35,7 @@ export default registerMPElement(`mp-drop-menu`, class extends Component {
         this.update({visibility: VISIBILITY_CLOSED});
         break;
       case VISIBILITY_OPEN:
+        this.dispatchEvent(new CustomEvent(`closing`));
```
jalack: my primary inclination was to do exactly that -- what made me wary was that I've seen callers in the past have code like the following `const open = ev.detail.state !== 'closed'`. obviously that's shoddy on the part of the callers, but I'm wary of breaking them all the same. if you think `close` is better than closing for a standalone event I'll rename.

lib/widgets/event-definition-tooltip/internal/index.styl
```diff
@@ -11,39 +11,35 @@
...
     opacity 0
     position absolute
-    transform translateX(-10px)
     width 340px
 
     &.in
       opacity 1
-      transform translateX(0)
-      transition all (mp-fast)ms ease-out-quart
+      transition all 100ms ease-out-quart
```
zakj: mp-just-so-fast?

lib/components/drop-menu/index.js
```diff
@@ -47,13 +49,19 @@ export default registerMPElement(`mp-drop-menu`, class extends Component {
...
         break;
       case VISIBILITY_CLOSED:
         this.update({visibility: VISIBILITY_OPENING});
+        this.dispatchCurrentVisibility();
         break;
     }
   }
 
+  dispatchCurrentVisibility() {
+    this.dispatchEvent(new CustomEvent(`state-change`, {detail: {state: this.state.visibility}}));
```
zakj: I'm not clear why we have both `state-change` and `change` custom events that seem to carry the same payload

lib/widgets/event-definition-tooltip/internal/index.styl
jalack: when `mp-fast` just ain't fast enough...

examples/style-guide-new/components/widgets/index.js
jbwyme: this is just fixture data isn't it?

lib/components/drop-menu/index.js
jbwyme: yeah why don't we just make the state-change the actual event? `this.dispatchEvent(new CustomEvent('opening'))`

lib/widgets/event-definition-tooltip/internal/index.js
jbwyme: IMO this seems like something the caller should handle, not the tooltip since this requires the tooltip to know about the implementation details of the menus

lib/components/drop-menu/index.js
jalack: apologies to @tdumitrescu if I'm paraphrasing him poorly, but per our discussion last night, his feeling was that nitty-gritty visibility changes don't warrant a top-level event (whereas `open` and `close` would). he preferred folding all such changes into a single event and letting callers look at event details to figure out the visibility state

lib/widgets/event-definition-tooltip/internal/index.js
jalack: the intent here is to interfere with the click handler in `mp-drop-menu` and stop the menu from closing. given that the click is registered on an element outside the `mp-event-definition-tooltip` (the actual element lives on the body), it would require a bunch of implementation-specific logic for callers to implement this effectively.

lib/components/drop-menu/index.js
jbwyme: I can understand that sentiment but it doesn't seem to me that the surface area of this component is going to fill up the global namespace of event names and it makes the API a lot more clear IMO. It also makes binding callbacks more explicit.

jbwyme: additionally, I've been meaning to add `open` and `close` as top level events anyways.

jbwyme: I'm attempting to convince @jalack to just use the existing `change` event and add two more calls with different `state`s. He's (rightfully) concerned that people have poor implementations that expect there to only be two states but I think we can check and see.

tdumitrescu: What I wanted was for `change` to remain the event that signals any of these different visibility state changes, not for there to be two events with subtly different names. I guess it's a matter of taste whether you think 'closing' and 'opening' should be top-level events; in which case i would argue that 'opened' and 'closed' should be the top-level events corresponding to the existing 'open/close' change events, but it's inconsistent with existing HTML event naming conventions and the whole thing seems shitty to me compared to keeping a change event for when this state changes.

jbwyme: see my comment above :)

PR #305, opened 2017-06-11
add icon for liveview

crmorford:


assets/icons/filter.svg
```diff
@@ -0,0 +1,14 @@
+<?xml version="1.0" encoding="UTF-8" standalone="no"?>
+<svg width="22px" height="20px" viewBox="0 0 22 20" version="1.1" xmlns="http://www.w3.org/2000/svg" xmlns:xlink="http://www.w3.org/1999/xlink">
```
jbwyme: icons need to be 22px x 22px to use the `svg-icon-style` mixin. You can modify the height and viewbox here and then use translate to move the icon down 1px so that it's centered (e.g. https://github.com/mixpanel/mixpanel-common/blob/master/assets/icons/hamburger.svg?short_path=7832713#L4)

PR #311, opened 2017-06-16
[smarthub] mp-smart-hub-tooltip + working report links

mduan:
- Create `mp-smart-hub-tooltip` (internally, it creates an element `mp-smart-hub-tooltip-internal` which is absolutely positioned relative to the document body)
- Update `smart-hub-alerts-data.json` with data to create real report links
- Some refactoring and bug fixes


lib/widgets/smart-hub-dropdown/index.js
```diff
@@ -24,6 +24,10 @@ export default registerMPElement(`mp-smart-hub-dropdown`, class extends Componen
           }
         },
         handleDropMenuChange: e => {
+          const state = e.detail.state;
+          if (![`open`, `closed`].includes(state)) {
```
rickydoar: in what cases would the state not be open or closed?

mduan: It can also be `opening` or `closing`. See https://github.com/mixpanel/mixpanel-common/blob/19b49203642aa4ebcc8fdc7ccd54cd49785a37f7/lib/components/drop-menu/index.js#L11-L14.

PR #315, opened 2017-06-22
Add mp drawer

yinfeiru:


examples/style-guide-new/components/components/index.js
```diff
@@ -8,6 +8,9 @@ import './index.styl';
 document.registerElement(`components-section`, class extends Component {
   get config() {
     return {
+      defaultState: {
+        drawerOpen: false,
```
tdumitrescu: looks like this ought to be added to the top-level 'open' field rather than declared here: https://github.com/mixpanel/mixpanel-common/blob/09b3c88b2878dffdcd40a4a5b68b69de66a4f19d/examples/style-guide-new/index.js#L34-L44

examples/style-guide-new/components/components/index.js
```diff
@@ -46,6 +49,7 @@ document.registerElement(`components-section`, class extends Component {
           }
           return match;
         },
+	shouldShowDrawer: drawerOpen => this.update({drawerOpen}),
```
tdumitrescu: "should show drawer" describes a boolean state; but this is a function that performs an action. follow the same pattern as the modals in this file: `openDrawer()`, `closeDrawer()` or potentially `toggleDrawer(true/false)`

examples/style-guide-new/components/components/index.jade
```diff
@@ -492,3 +492,44 @@
...
                 br
                 span.mp-red     mp-spinner
+
+  if sectionOpen == 'components' || searchSections.includes('drawer')
+    .sub-section-container#drawer
+      .divider
+      .section-content
+        section
+        .subsection-title Drawer
+        .subsection-description For displaying modals that slide in from the side.
```
tdumitrescu: 'modal' seems like an odd description of a drawer. @EricBenjamin do you have reactions?

examples/style-guide-new/components/components/index.jade
```diff
@@ -492,3 +492,44 @@
...
+        .indent-container
+          .example-block
+            mp-button.mp-button-secondary(on={click: () => $helpers.shouldShowDrawer(true)}) Show drawer
```
tdumitrescu: `on={click: openDrawer}` boom, done!

examples/style-guide-new/search-terms.json
```diff
@@ -201,7 +201,16 @@
...
   ],
-  "ui-icons": [
+  "drawer": [
+    "components",
+    "mp-drawer",
+    "drawer",
+    "left",
+    "right",
+    "up",
+    "down"
```
tdumitrescu: what do these terms have to do with the drawer component?

lib/components/drawer/index.jade
```diff
@@ -0,0 +1,12 @@
+.mp-drawer-stage
+  if !$component.isAttributeEnabled(`hide`)
```
tdumitrescu: i _think_ (via @jbwyme) that we have by now settled on using the attribute `open` to control visibility of components like this. At the very least, having a positive rather than negative attr (open/show rather than closed/hide) avoids the possibility of temporary content flash during component initialization if attrs haven't yet been added (i.e. the default behavior of "no attribute present" should be to keep the component hidden).

lib/components/drawer/index.js
```diff
@@ -0,0 +1,44 @@
...
+
+document.registerElement(`mp-drawer`, class extends Component {
+  get config() {
+    return {
+      css,
+      template,
+      useShadowDom: true,
+
+      defaultState: {
+        delayRemove: 250, // TODO replace with var once in common
```
tdumitrescu: wanna DO this TODO? this is `mp-fast`: https://github.com/mixpanel/mixpanel-common/blob/09b3c88b2878dffdcd40a4a5b68b69de66a4f19d/lib/stylesheets/mixins/timing.json

lib/components/drawer/index.js
```diff
@@ -0,0 +1,44 @@
...
+
+  attributeChangedCallback() {
+    super.attributeChangedCallback(...arguments);
+
+    Array.from(document.querySelectorAll(`body, html`))
+      .forEach(el => {
+        el.style.overflow = this.isAttributeEnabled(`hide`) ? `` : `hidden`;
+      });
+
+    this.update({placement: this.getAttribute(`placement`) || `right`});
```
tdumitrescu: do we actually have CSS implementations of drawers in different positions? ie shall we just get rid of this attr until it is needed?

lib/components/drawer/index.styl
```diff
@@ -0,0 +1,73 @@
...
+      transform translate3d(0%, 0, 0)
+
+  &.mp-drawer-out
+    .mp-drawer-overlay
+      opacity 0
+
+    .mp-drawer-bg, .mp-drawer-content
+      transform translate3d(100%, 0, 0)
+
+  &.mp-drawer-placement-top
```
tdumitrescu: the only implementation right is top? like, coming down from the top of the page?

lib/components/drawer/index.js
```diff
@@ -0,0 +1,44 @@
...
+      },
+    };
+  }
+
+  attachedCallback() {
+    if (this.initialized) {
+      return;
+    }
+
+    if (!this.initialized) {
```
tdumitrescu: why the conditional? if you got to this line, then `this.initialized` is definitely false.

lib/components/drawer/index.styl
yinfeiru: Yes, the drawer might coming down from the top. By default, it's from the right hand side.

lib/components/drawer/index.js
yinfeiru: yup 
`&.mp-drawer-placement-top`

lib/components/drawer/index.js
yinfeiru: I'll double check this logic.

examples/style-guide-new/search-terms.json
yinfeiru: Ah.. Maybe I was overthinking about this but since this drawer can slide in from either the top or the right hand side, the 4 key words above all sound relevant to me - "slide in to the left" "slide in from the right".. I don't really know if they are needed.

tdumitrescu: I'd say if you're going to have that it's fine; but you actually have to implement them! and provide examples from the different sides in the styleguide, since they can all interact with the page in pretty different ways.

examples/style-guide-new/components/components/index.js
yinfeiru: ab031b8

examples/style-guide-new/components/components/index.js
yinfeiru: good point. I'll do it with `toggleDrawer` since both `openDrawer` and `closeDrawer` are just toggling the visibility with the same code.
ab031b8

examples/style-guide-new/components/components/index.jade
yinfeiru: It's handled in the commit above.

examples/style-guide-new/search-terms.json
yinfeiru: I'll just remove those key words to not expending the scope. e1c10f1

lib/components/drawer/index.jade
yinfeiru: 813548b

lib/components/drawer/index.js
yinfeiru: TODO -> DONE
277447d

examples/style-guide-new/components/components/index.jade
EricBenjamin: hrmm, interesting... 🤔 It has many similarities to our modal definition: shade, new UI, etc. I think we should call this a "panel" – if we're just fussing over the desc. I can pull, edit, and PR. I'm wondering if there should be a min width on this to force consistency when content isn't necessarily the same.

lib/components/drawer/index.styl
EricBenjamin: Ah, see. this isn't JUST the bookmarks panel. 👍

examples/style-guide-new/components/components/index.jade
yinfeiru: f5f0c7a

@EricBenjamin Let me know if you want to add a min-height/width to it. I'd like to skip adding it just to unblock myself. But if you think it's an important thing to do, I'm fine with it too.

yinfeiru: 2957849
chatted with @EricBenjamin offline, we want to set `min-width` to `690px` when it comes in from the right hand side.

lib/components/drawer/index.js
```diff
@@ -0,0 +1,44 @@
...
+    super.attachedCallback(...arguments);
+  }
+
+  attributeChangedCallback() {
+    super.attributeChangedCallback(...arguments);
+
+    Array.from(document.querySelectorAll(`body, html`))
+      .forEach(el => {
+        el.style.overflow = this.isAttributeEnabled(`open`) ? `hidden` : ``;
+      });
```
zakj: this feels like the sort of thing that's very confusing to debug later ("why does body suddenly have overflow hidden? I'll just grep for `overflow: hidden`...", or even more tricky, "I explicitly set body overflow style to `scroll` but now it's unset, what happened?", etc.). a better option might be to add this to `body.no-scroll` in our default stylesheet and just toggle the class based on whether the drawer is open.

tdumitrescu: sounds ok to add to https://github.com/mixpanel/mixpanel-common/blob/master/lib/stylesheets/defaults/index.styl, though it would make this the only component besides `svg-icon` that is not self-contained.

examples/style-guide-new/components/components/index.jade
```diff
@@ -492,3 +492,44 @@
...
                 br
                 span.mp-red     mp-spinner
+
+  if sectionOpen == 'components' || searchSections.includes('drawer')
+    .sub-section-container#drawer
+      .divider
+      .section-content
+        section
+        .subsection-title Drawer
+        .subsection-description For displaying panels that slide in from the side.
```
tdumitrescu: from the side...or top?

examples/style-guide-new/components/components/index.jade
```diff
@@ -492,3 +492,44 @@
...
+        .indent-container
+          .example-block
+            mp-button.mp-button-secondary(on={click: () => $helpers.toggleDrawer(true)}) Show drawer
```
tdumitrescu: You've got implementations of side and top, so please add an example here of top-placement. Each of top, bottom, left, right has its own peculiarities in terms of how it interacts with the rest of the page/viewport, so we need a sample of each one that's implemented.

examples/style-guide-new/components/components/index.js
```diff
@@ -46,6 +46,7 @@ document.registerElement(`components-section`, class extends Component {
           }
           return match;
         },
+	      toggleDrawer: drawerOpen => this.update({drawerOpen}),
```
tdumitrescu: - indentation is weird
- my previous suggestion was to add this to the `open` entry in state, not to make it another top-level entry
- FWIW I think you should do what's consistent with the rest of this file, i.e. `openDrawer()` and `closeDrawer()`; or else make those consistent with your new way

lib/components/drawer/index.jade
```diff
@@ -0,0 +1,12 @@
+.mp-drawer-stage
+  if $component.isAttributeEnabled(`open`)
+    .mp-drawer(class={
+      delayed: {'mp-drawer-in': true},
+      remove: {'mp-drawer-out': true, delayRemove},
+      [`mp-drawer-placement-${placement}`]: true,
+      'mp-modal-absolute': $component.isAttributeEnabled('not-fullscreen')
```
tdumitrescu: why does the drawer component have a class `mp-modal-absolute`?

lib/components/drawer/index.js
```diff
@@ -0,0 +1,44 @@
...
+      helpers: {
+        close: () => this.dispatchEvent(new CustomEvent(`close`)),
+      },
+    };
+  }
+
+  attachedCallback() {
+    if (this.initialized) {
+      return;
+    }
```
tdumitrescu: This is [already in master in `panel`](https://github.com/mixpanel/panel/blob/89b7cabbd32d6d1231ce554a1e8590e4e4ae32ab/lib/component.js#L215-L218) so this component doesn't need to override `attachedCallback` at all.

lib/components/drawer/index.js
tdumitrescu: also re: debugging, it shouldn't be hard to answer the question "why does body suddenly have overflow hidden?" given that this only happens when the drawer is open, prompting the body to get covered by a semi-opaque overlay.

lib/components/drawer/index.styl
```diff
@@ -0,0 +1,76 @@
+@import '../../stylesheets/defaults'
+
+.mp-drawer
+  height 0
+  overflow hidden
+  position fixed
+  right 0
+  top 0
+  width 0
+  z-index 100
```
tdumitrescu: we have variables now: https://github.com/mixpanel/mixpanel-common/blob/e1ae685e8b125e5cb38cbcb66608b07782e1e8d9/lib/stylesheets/mixins/common.styl#L11-L15. @jbwyme should this be `$z-modal`?

lib/components/drawer/index.js
zakj: I'd argue any component that queries DOM outside its shadow root is already not self-contained (though I take your point about requiring external styles), but regardless I'd like to minimize how we affect elements a component doesn't own

examples/style-guide-new/components/components/index.jade
yinfeiru: edfe5f6

examples/style-guide-new/components/components/index.js
yinfeiru: It's handled in the commit above (since they are related)

lib/components/drawer/index.jade
yinfeiru: so you can choose not to make it cover the whole screen
https://app.zeplin.io/project/5810e4117899d27c64470e4d/screen/591f1e2635cbfd15f1c0b315

lib/components/drawer/index.js
yinfeiru: d30f28b

examples/style-guide-new/components/components/index.jade
yinfeiru: 2c31d3e

lib/components/drawer/index.styl
yinfeiru: 8cdf699
let me know if this is not the value i should use @jbwyme

lib/components/drawer/index.jade
tdumitrescu: then call it `mp-drawer-fullscreen` (my question was about it referring to 'mp-modal' which is a different component type). also looking at that design, that attribute probably needs to affect more than just that little bit of css; for instance there's no reason for it to mess with `body` or `html` if it's not a fullscreen drawer.

yinfeiru: yea, good point. 
3c7ca22
b580e26

In terms of whether or not more css changes will be needed, I'm actually not that sure at this point.

Maybe I can just import `mixpanel-common/build/index`? I'll try it out.

PR #317, opened 2017-06-26
Add stylesheet for tables

jalack:
Adding a stylesheet for tables so that we can have consistent table styles across Lexicon, Notifications, and Cohorts without sacrificing too much flexibility. I'm open to changes in the directory structure, but the standalone, non-composable nature of table styles led me to separate it from mixins.


tdumitrescu:
published 0.96.3

PR #319, opened 2017-06-29
SST bug fixes

nojvek:
Fix CA-2874: Caret size change during property pill hover state
Fix 2824: Long event names get cut off
Fix CA-2861: Add 'SFDC data only' copy to Filter dropdown

image


lib/widgets/items-menu/index.jade
```diff
@@ -9,6 +9,8 @@
     else if emptyMessage
       li.list-option.list-empty= emptyMessage
   else
+    if datasetDisplayName
+      .dataset-only-label= `${datasetDisplayName} data only`
```
evnp: What would you think about making this more generic, like `hint-text` or something? Not sure how I feel about datasets leaking into mixpanel-common. @jbwyme thoughts?

tdumitrescu: absolutely - this is a generic UI component for displaying a list of items for a menu. this looks like it can be treated as some kind of additional filtering and/or labeling, without getting into the specifics of 'datasets'

nojvek: Makes sense. I'm calling it `sectionsHint` if that works with you all.

tdumitrescu:
published 0.98.0

PR #320, opened 2017-06-29
Add table styles for sort carets

jalack:
- Add styles for sort carets
- Remove border between titles
- Header font-size is 14px, not 16px

screen shot 2017-06-29 at 11 04 29 am


jayjo:
Header font-size is 13px in all of the mocks and proposals - when did that change to 14px?

jalack:
@EricBenjamin and Stephanie indicated 14px to me, not sure when it changed

lib/stylesheets/components/mp-table.styl
```diff
@@ -15,27 +15,32 @@ table-cell-padding = 22px
...
+    &:not(.sort-column)
+      .sort-controls
+        visibility hidden
+
+    &.can-sort:hover
+      cursor pointer
+      .sort-controls
+        visibility visible
 
   .body
```
jbwyme: seems like we should be using prefixing for a lot of these no? I could see `body`, `titles`, `cell`, etc being overridden or accidentally being matched by other random css.

jalack: we're using the prefix `mp-table` on [line 13](https://github.com/mixpanel/mixpanel-common/blob/98390932a4ca4a382b3b4dbede84635d2d1ee433/lib/stylesheets/components/mp-table.styl#L13)

jbwyme: oh sick

